PR #39, opened 2018-05-02
Fix typo in directory name

shailysangwan:
As seen in https://github.com/frappe/accounting/blob/master/reports/index.js#L2

    const GeneralLedger = require('./GeneralLedger/GeneralLedger')


shailysangwan:
Sorry, didn't notice that there was an existing PR for this.

PR #40, opened 2018-05-05
Fix start script

shailysangwan:


shailysangwan:
Fixed in https://github.com/frappe/accounting/commit/74b1c49845088db7ceeca48bb7c0dcd86d9f43c5

PR #43, opened 2018-06-14
Initialize UOM model and create Warehouse model

jaffreyjoy:
#### The `UOM` model was initialized with the fields:
 - `itemname`  (Name of the Item) 
 - `uom` (Unit of Measure for the item)

#### The `Warehouse` model was created with the fields:
 - `name` (Warehouse name)
 - `parentWarehouse` (Name of parent warehouse (if exists) (else create one)
 - ` isGroup` (flag to identify whether warehouse is of type `group` or not)

> The `Warehouse` model has a functioning `tree view` that is generated by the desk and a `list view` to edit or delete the corresponding warehouse entry.

> The list view has the `warehouseName`, `type` ( i.e. `child` or `group` ) and if it has a parent warehouse  then it's `parentWarehouse` is mentioned.

Here is the GIF of the working of the `Warehouse` model:
warehouse model working GIF

 - GIF [link](https://kek.gg/i/4PVb6B.gif) (if at all the embedded one seems somewhat unclear 😅)
 - If the GIF is too fast here's the video [link](https://vimeo.com/275105013)

@rmehta @netchampfaris Please review and mention any changes to be done 🙏 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,12 +23,7 @@
     "build": "node build/build.js"
   },
   "dependencies": {
-    "bootstrap": "^4.1.1",
-    "feather-icons": "^4.7.3",
-    "frappejs": "^0.0.8",
-    "vue": "^2.5.2",
-    "vue-flatpickr-component": "^7.0.2",
-    "vue-router": "^3.0.1"
+    "frappejs": "^0.0.8"
   },
   "devDependencies": {
     "autoprefixer": "^7.1.2",


client/index.js
```diff
@@ -10,6 +10,8 @@ module.exports = {
 
         registerReportRoutes();
 
+        frappe.desk.menu.addItem('Uom', '#list/Uom');
+	    frappe.desk.menu.addItem('Warehouse', '#tree/Warehouse');
```
rmehta: check spacing

models/doctype/Uom/Uom.js
```diff
@@ -0,0 +1,36 @@
+module.exports = {
```
rmehta: How do you use this UOM?

UOM does not have an item property. They have a many-to-many relationship, so `Item UOM` should be a separate DocType

models/doctype/Uom/Uom.js
```diff
@@ -0,0 +1,36 @@
...
+    name: "Uom",
+    doctype: "DocType",
+    naming: "random",
```
rmehta: Naming of UOM should not be random! Should be the `name` property. See Item or Party

models/doctype/Warehouse/Warehouse.js
```diff
@@ -0,0 +1,55 @@
+module.exports = {
```
rmehta: check indentation

rmehta (changes requested):

jaffreyjoy:
### Updated UOM model
 - Has single field `name`
 - Link created in Item model

### Working ⬇️
 uom-updated-GIF

 - GIF [link](https://kek.gg/i/5YSwyP.gif)
 - Video [link](https://vimeo.com/275486629)

@rmehta @netchampfaris Please review and mention any changes to be done. :pray:

PR #44, opened 2018-06-15
Email Module Milestone #1

inishchith:
## Email Module 

- [x] Sending Emails
- [x] Receiving Emails 
- [x] Syncing 

### Future Work
- [TODO](https://github.com/email-frappejs/accounting/issues/7)

### [Demo](https://vimeo.com/275231854)

@rmehta @netchampfaris @surajshetty3416 
Sir , Please let use know if any changes required 😄


email/getConfigReceiver.js
```diff
@@ -0,0 +1,15 @@
+const frappe = require('frappejs');
```
surajshetty3416: You can commonify the code to get receiver and sender account config.... you can pass filters accordingly

email/receiver.js
```diff
@@ -0,0 +1,76 @@
+const frappe = require('frappejs');
+const simpleParser = require('mailparser').simpleParser;
+const Imap = require('imap'),
+  inspect = require('util').inspect;
```
surajshetty3416: `inspect` Not used in the code

models/doctype/Email/Email.js
```diff
@@ -0,0 +1,84 @@
+const frappe = require('frappejs');
```
surajshetty3416: commented code

models/doctype/Email/EmailForm.js
```diff
@@ -0,0 +1,22 @@
...
+
+    makeToolbar() {
+        this.sendButton = this.container.addButton(frappe._("Send"), 'primary', async (event) => {
+           frappe.call({
+               method: 'send-mail',
+               args: this.doc.getValidDict()
+           });
+        });
+        this.on('change', () => {
+            const show = this.doc._dirty && !this.doc.submitted;
```
surajshetty3416: Recheck if this on change listener is correctly implemented or needed (?)

inishchith:
@surajshetty3416 Will make the necessary changes and Get back !

models/doctype/Email/EmailForm.js
inishchith: Gotcha 🙈  . Resolved! . now we aren't explicitly doing an `insert` . 
`await this.save` does the job .

inishchith: @surajshetty3416 only doubt . `this.doc.name` in this case will be some random string . i want to assign it to the return value of the function call  (`frappe.call` thing) . i tried this doesn't work 🙈

email/getConfigReceiver.js
inishchith: Done . 
i didn't use filters as we need to work on another issue regarding `defaultIncoming` and `defaultOutgoing` , so that incorporating those changes would take less time IMO .

inishchith:
@rmehta @surajshetty3416 @netchampfaris i've made the necessary changes . let us know if there's any more to be Done 😄

netchampfaris:
Great work 👍 

Since, this PR relies on the old UI, let's wait until it is compatible with the new UI based on Vue before merging.

inishchith:
@netchampfaris 👍  we're currently working on E-mail view on `move-to-vue` branch .

inishchith:
this is closed in reference to an updated one #50

PR #143, opened 2020-05-12
Update locale and currency information for Curacao

frankgeerlings:
This change adds Curaçao's currency and time zone (as listed in IANA time zone db).

### DIFF
--- a/fixtures/countryInfo.json
+++ b/fixtures/countryInfo.json
@@ -677,10 +677,15 @@
     },
     "Cura\u00e7ao": {
         "code": "cw",
+        "currency": "ANG",
         "currency_fraction": "Cent",
         "currency_fraction_units": 100,
+        "currency_name": "Netherlands Antillean Guilder",
         "currency_symbol": "\u0192",
-        "number_format": "#,###.##"
+        "number_format": "#,###.##",
+        "timezones": [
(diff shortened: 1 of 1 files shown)


piyushsinghania (commented):
I believe the changes are legitimate and can be merged. 

One thing I noticed is that the currency symbol "\u0192" doesn't appear in the application. The currency symbol is kind of small ' f '. @frankgeerlings can you look into this, if possible?

frankgeerlings:
At the time the pull request was made the ƒ, a stylized 'f' (abbreviation for 'florin') was already present and did appear in the application.

As you can see in the diff it was already there. I would say the issue you raise would be beyond the scope of this request. Should perhaps be a new issue?

piyushsinghania:
@frankgeerlings I think you are correct, as of now, (when making similar changes as of yours) and then using "ANG" as the currency, I am not able to get the florin ƒ symbol. Maybe it's a new issue, can you help me by confirming it?

18alantom:
Screen Shot 2021-10-21 at 11 53 36

Merging it as the Florin symbol appears after a restart. This may be the case for all currencies check it for USD (non-Unicode def) and INR (Unicode def). PR is flawless.

PR #152, opened 2020-06-10
fix: Dynamic Link Target

ruchamahabal:
Target undefined for the Dynamic Link field.

**Before:**

create-new-payment

**After:**

payment-works


18alantom:
PR seems proper; Travis is failing for unrelated reasons. Merging.

PR #153, opened 2020-06-10
fix: formula for amount field in Payment

ruchamahabal:
Payment table creation failing during setup:

db

After setting formula for amount field in Payment, the setup works:

solved


18alantom:
Closing as this was included in #220.

PR #211, opened 2021-08-18
Fix reported issues and imporvements

ankitsinghaniyaz:


18alantom:
Closing because #220 and #216 contain these commits.

PR #222, opened 2021-10-21
Resolve ledger related issues

ankitsinghaniyaz:
- add the ability to filter out the reverted ledger entries
- fix issues with ledger entries not created when doing partial payments

### DIFF
--- a/reports/GeneralLedger/viewConfig.js
+++ b/reports/GeneralLedger/viewConfig.js
@@ -56,6 +56,18 @@ const viewConfig = {
       placeholder: 'To Date',
       label: 'To Date',
       fieldname: 'toDate'
+    },
+    {
+      fieldtype: 'Select',
+      options: [
+        { label: 'Hide Audits', value: '1' },
+        { label: 'Show Audits', value: '' }
+      ],
+      size: 'small',
+      label: 'Audits',
+      placeholder: 'Show Audits',
+      default: 1,
+      fieldname: 'reverted'
     }
   ],
   method: 'general-ledger',
--- a/reports/GeneralLedger/viewConfig.js
+++ b/reports/GeneralLedger/viewConfig.js
@@ -59,11 +59,13 @@ const viewConfig = {
     },
     {
       fieldtype: 'Select',
-      options: ['Show Audits', 'Hide Audits'],
+      options: [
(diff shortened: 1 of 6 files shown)


reports/GeneralLedger/viewConfig.js
```diff
@@ -56,6 +56,18 @@ const viewConfig = {
...
+      options: [
+        { label: 'Hide Audits', value: '1' },
+        { label: 'Show Audits', value: '' }
+      ],
+      size: 'small',
+      label: 'Audits',
+      placeholder: 'Show Audits',
+      default: 1,
+      fieldname: 'reverted'
     }
```
18alantom: This should be a _Check_ `fieldtype` called "Show Reverted" with the default as 0.

Reasoning: choice here is binary so _Select_ is not really required. Also when _Select_ is used 2 clicks are required, as opposed to 1 when _Check_ is used.

ps: I know the check doesn't look good, it defaults to native we'll have to fix its design.

piyushsinghania: @18alantom I really appreciate your attention to detail ("when _Select_ is used 2 clicks are required, as opposed to 1 when _Check_ is used"). 

We will update it as you have suggested.

First, we had used _Check_ only. And just as you said, its design doesn't fit well there. So, we moved to _Select_.

ankitsinghaniyaz: @18alantom I read your conversation regarding making this into a checkbox and that makes sense and like you said it's looking really bad. Without design help, I'm not sure if we will be able to fit in the checkbox in the top bar. Maybe you can give that a shot?

Few more pointers I wanted to add:
1. I would like if we can use the word `audit` instead of `reverted` as a common user will be confused about what `show reverted` would mean in this context easily. We can have a chat around this on our Telegram and decide.
2. Another point is about the reverted field being a binary, which I think will not work. Because as a user we want to see either all the entries or only non-reverted entries and here's how they map to different values.

- reverted === 0: non-reverted entries
- reverted === 1: reverted-entries
- reverted === nil: all entries

This works easier with select, could also work with a checkbox but need some design help there.

cc: @piyushsinghania

18alantom:
Merging; everything works as stated; binary selects will be updated to a _Switch_ UI element later on.

PR #230, opened 2021-10-30
Add validation when paying excess amount

ankitsinghaniyaz:


models/doctype/Payment/Payment.js
```diff
@@ -102,7 +103,17 @@ module.exports = {
...
-      formula: doc => doc.getSum('for', 'amount')
+      formula: doc => doc.getSum('for', 'amount'),
+      validate(value, doc) {
+        const amount = doc.getSum('for', 'amount');
+        const isValid = amount >= value && value > 0;
+
+        if (!isValid) {
+          throw new frappe.errors.ValidationError(
+            `Payment amount ${value} cannot be accepted. Please enter a value between 0 and ${amount}`
+          );
```
18alantom: @piyushsinghania on checking the change, it doesn't allow me to set a value greater than what is expected:

https://user-images.githubusercontent.com/29507195/139625627-fb91d197-0bb8-47e6-8407-f624bb7ebab0.mov

So keeping that in mind I suggest a few changes:
1. Reword the message to something like _"Payment amount cannot exceed ₹ ${amount}."_ , or you can reword it to also account for the fact that the value has been reset to the max viable amount (in haste they may not notice this, and later on it can be an issue).
2. And _"Payment amount cannot be ₹ 0.00"_ in case the amount is 0.
3. Don't forget the currency sigil. For this, we could use a helper function to format currencies to the correct precision with the correct sigil.
4. Don't forget to wrap the message in `frappe._` to support translations later when we add them.

Other than that it works perfectly, I think capping it to the max viable amount is a nice touch!

(cc: @ankitsinghaniyaz)

ankitsinghaniyaz: #1 we'll verify and 100% on the other points.

ankitsinghaniyaz:
@18alantom addressed all the feedback

models/doctype/Payment/Payment.js
```diff
@@ -102,7 +103,27 @@ module.exports = {
       label: 'Amount',
       fieldtype: 'Currency',
       required: 1,
-      formula: doc => doc.getSum('for', 'amount')
+      formula: doc => doc.getSum('for', 'amount'),
+      validate(value, doc) {
+        const amount = doc.getSum('for', 'amount');
+
+        if (value > amount)) {
```
ankitsinghaniyaz:
```suggestion
        if (value > amount) {
```

18alantom:
Didn't realize the capping was auto 😅
Either way, better than spamming with errors until change.

PR #234, opened 2021-11-03
Fix profit and loss graph not reflecting changes correctly

ankitsinghaniyaz:


18alantom:
P&E, Expenses are better than before, but still requires a lot of polish.

Issue #141, opened 2020-05-10
multiple tax rates in invoices should show as separate line items


test2a:
https://i.imgur.com/RlNEMKN.png

attached screenshot where i set one item as 18% and another as 28%. the software seems to calculate proper tax but only shows the tax classification of the first item and adds taxes of all there. it should be shown separately.
add: minor typo, in case of "sales", it shows "make payment". it should read "receive payment"
https://i.imgur.com/Qza1V9C.png

test2a:
second, is the team working on indian taxation because i am seeing issues. would it be okay to list them out separately or at once or the team wants to work on bigger issues right now?

18alantom:
@test2a we have recently restarted development on Books. If you are still interested you can list out the issues with Indian taxation on a _New Issue_.

18alantom:
Will add per-item tax breakdown when invoice item row expansion is added, for now incorrect display of rates has been removed.

PR #245, opened 2021-11-11
fix(ux): remove tax rate from invoice breakdown

18alantom:
Removed display of tax rate from the tax breakdown section.

**Why**
- One would be more concerned with how much is paid for each tax.
- Item-wise breakdown on the invoice level is cluttersome.
- This information can be displayed at the item level, will add this later by allowing row expansion.

(created a PR to add the above message)


18alantom:
Renders https://github.com/frappe/books/issues/141 redundant.

PR #247, opened 2021-11-14
Fix error messgae on navigating to dashboard

ankitsinghaniyaz:


ankitsinghaniyaz:
The current solution will suppress all the errors which we don't want. Maybe try this solution instead?
https://stackoverflow.com/a/67102038/3541205

18alantom:
Instead of catching the error why not prevent push if the current route is the same?

(cc: @piyushsinghania )

piyushsinghania:
@18alantom you are right.
I have made the suggested changes.

18alantom:
@piyushsinghania  it's working properly 👍

I think you should create a separate function for routing in the [utils](https://github.com/frappe/books/blob/master/src/utils.js) file, cause there are other places (like [this](https://github.com/frappe/books/blob/master/src/App.vue#L96)) where the error pops up.

We can then use that function for all our routing purposes.

(_I should've probably mentioned it earlier, sorry bout that_)

piyushsinghania:
@18alantom that sounds good, I will do that.

piyushsinghania:
@18alantom I have separated the function to utils and used the same at possible places.

18alantom:
@piyushsinghania looks good to me; just one thing, can you check why diffs from [this](https://github.com/frappe/books/commit/aa6a565f9b227cfe9488c0e6e71cc3346f2ef260) commit are showing in this PR?

Issue #235, opened 2021-11-05
Fix the Dashboard. Profit & Loss, Top Expenses Section


18alantom:
- [x] Expanding the books window resizes the dashboard, but compressing it doesn't and require scrolling.
- [x] Resizing messes up the tooltip location.
- [x] Donut chart hovering is a bit messed up.
- [x] Switching periods sometimes doesn't load the correct data (after removing the Selector issue PR # 234).
- [x] Dashboard Refactor
  - [x] Cashflow Line
  - [x] Profit and Loss Bar
  - [x] Top Expenses Donut
- [x]  If data not there in a period then the period selector disappears, this requires a refresh to fix (# 234).
- [x] Vertical and horizontal alignment.
- [x] Layout doesn't match the invoice section above.
- [x] Expenses UI seems broken, text overlays the pie chart.
- [x] uncaught Reference transform not defined on hovering the pie chart.
- [x] uncaught DOMEx when changing the donut period.
- [x] Text in the donut chart isn't centered.

18alantom:
Will be deprecating the use of Frappe Charts and rewriting custom chart logic using `svgs`. 

Reasons for this:
- Frappe Charts has hardcoded offset values that require hacky CSS to fix.
-  Some behaviour is unpredictable such as the transform error on hovering over the donut charts.
- Tooltip movement is jarring.
- Doesn't have (afaik) any hover animations such as the donut chart scaling.
- Chart resizing doesn't take place on window resize. (expansion works, compression doesn't)

This will take some time but should be done before version 1.0

18alantom:
The dashboard issues have been fixed to a large extent, I'm still unhappy with how it looks. But fixing that will require a redesign.

PR #262, opened 2021-11-29
Centralised the expenses in the expenses chart

piyushsinghania:
image


18alantom:
@piyushsinghania
 
This doesn't fix the underlying issue, i.e. the text isn't constrained by the donut chart, they are both sibling divs. Imo that entire section requires a refactor 😔

Screen Shot 2021-11-29 at 18 10 11

piyushsinghania:
@18alantom you are absolutely right.
Can we add this to frappe books board,
we will refactor the component/Section.

18alantom:
@piyushsinghania I've renamed [#235](https://github.com/frappe/books/issues/235). If you have any points regarding the dashboard fixup do add it to that issue.

Issue #185, opened 2020-12-28
Option for precision


manomayholdings:
Hello Contributors,
I Have Found That During Product Add Process My Product Price is 0.567 Rs Per Item So Book rounded Off to 0.58 Rs But as No Of Product is 120000 So 0.567 - 0.58 = 0.013
So 0.013 X 120000 = 1560 Difference*

So Can You Share Me How We Can Add Amount more than 2 Decimal or As Per Choice For All Books.

THANKS
![0 576](https://user-images.githubusercontent.com/25424914/103221645-1f35ec80-4949-11eb-90f4-5e2a83c88703.png)

piyushsinghania:
I appreciate your observation. And I would like to add up by asking whether you entered 0.576 or 0.567 in the Rate section? As I can see in the above image you might have entered the wrong rate (In the section where you are adding the rate of your product, I see it to be 0.576, which if rounded is 0.58). While in the calculations above you are using the rate of the product as 0.567 and hence a difference of 0.013 is arising rather than 0.003 which in turn would have reduced the difference significantly.

In my mind it can be fixed by changing the decimal points to get the desired value.

18alantom:
Books rounds of values to the second decimal place, need to add a _precision_ setting.

18alantom:
Closing cause: https://github.com/frappe/books/pull/268 merged

PR #268, opened 2021-12-02
Refactor currency handling to use pesa

18alantom:
### Currency Refactor
PR pertains to refactoring books to use [`pesa`](https://github.com/frappe/pesa) to handle monetary calculations.

#### Pesa initialization
1. Initialized immediately after db connect.
2. If setup is incomplete currency is not available; ∴ initialized with default `"XXX"` code for no currency.
3. if setup incomplete, on submit with the entered currency.

#### Todo List
- [x] All javascript calculations should take place through `pesa` methods.
- [x] All sql queries should cast the values to number first.
- [x] Deprecate earlier handling of precision.
- [x] Setting display should change the number of significant digits displayed.
- [x] On storing the values in the db, use string type and use `Money#store`
    which will get a precision adjusted decimal string.
- [x] When a doc is being fetched the currency values should be converted to the
    Money object.
- [x] Patch that converts values to store strings.
- [x] After db init when `currency` and `internalPrecision` are available set `pesa` on
    the global object.

#### Other Changes
- ~`Model#field#validate(value, doc)` will be called before update in `Base.vue` on `blur`ring of the input field~. Validations will be inline, previous change has been removed cause they are called on `doc.set`.

https://user-images.githubusercontent.com/29507195/148735790-eb3cca97-a146-4354-9a55-47760c030eda.mov

- This function should throw an error if invalid, this will be show in a message box.
- `Int` and `Float` types now have number inputs, `Model#field#minValue` and `Model#field#maxValue` can be used to restrict the range.
- Added **Locale** in _System Settings_ to set `Intl.NumberFormat`

FrappeJS: https://github.com/frappe/frappejs/pull/153

### DIFF
--- a/models/doctype/Payment/PaymentServer.js
+++ b/models/doctype/Payment/PaymentServer.js
@@ -46,9 +46,9 @@ export default class PaymentServer extends BaseDocument {
   }
 
   updateAmountOnReferenceUpdate() {
-    this.amount = 0;
+    this.amount = frappe.pesa(0);
     for (let paymentReference of this.for) {
-      this.amount += paymentReference.amount;
+      this.amount = this.amount.add(paymentReference.amount);
     }
   }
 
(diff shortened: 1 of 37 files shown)


18alantom:
All builds are gonna fail cause they're dependent on the `make-money` branch of `frappejs`.

models/doctype/Payment/PaymentServer.js
```diff
@@ -113,18 +114,28 @@ export default class PaymentServer extends BaseDocument {
...
+            'Currency'
+          )}.`
         );
-        if (this.amount <= 0) {
-          const amt = this.amount < 0 ? ` (${this.amount})` : '';
+
+        if (this.amount.lte(0)) {
+          const amt = this.amount.lt(0)
+            ? `: ${frappe.format(this.amount, 'Currency')}`
+            : '';
```
ankitsinghaniyaz: Seems like with pesa the ternary comparison wont be required.

```suggestion

        if (this.amount.lte(0)) {
          const amt = ${frappe.format(this.amount, 'Currency')};
```

src/components/Controls/Int.vue
```diff
@@ -4,11 +4,16 @@ import Data from './Data';
 export default {
   name: 'Int',
   extends: Data,
+  computed: {
+    inputType() {
+      return 'number';
+    },
```
ankitsinghaniyaz, line 10: The new number inputs have added the up/down arrows for the fields, They seem less useful to me and we can hide them through css.

18alantom:
Merging this should close #284 and #185

ankitsinghaniyaz (approved):

PR #277, opened 2021-12-14
Add GSTR-I reports for B2B

ankitsinghaniyaz:


accounting/gst.js
```diff
@@ -0,0 +1,197 @@
...
+  // based condition we need to triggered different methods
+  if (transferType === 'B2B') {
+    gstData.b2b = await generateB2bData(report.rows);
+  } else if (transferType === 'B2CL') {
+    gstData.b2cl = await generateB2clData(report.rows);
+  } else if (transferType === 'B2CS') {
+    gstData.b2cs = await generateB2csData(report.rows);
+  }
+
+  await sleep(1);
```
18alantom, line 97: Any reason for a 1 second sleep?

package.json
```diff
@@ -25,6 +25,7 @@
     "core-js": "^3.19.0",
     "csvjson-csv2json": "^5.0.6",
     "electron-store": "^8.0.1",
+    "file-saver": "^2.0.5",
```
18alantom: This doesn't seemed to be used anywhere, saving seems to use `fs`. I guess it can it be removed?

src/saveReportAsJson.js
```diff
@@ -0,0 +1,9 @@
+import fs from 'fs';
+import { shell } from 'electron';
+
+export default async function makeJSON(data, savePath) {
+  fs.writeFile(savePath, data, (error) => {
+    if (error) throw error;
+    return shell.openPath(savePath);
```
18alantom: Opening `.json` by default will in most cases just launch the text editor, raw json isn't something people generally would want to look at other than to verify that it did get saved.

I feel we should do one of these:
1. Open the folder where it's saved.
2. Show a toast saying that the file was saved at whatever path and clicking the toast opens the folder.

I don't know if we have UI for a toast, if we don't you can go with 1.

src/sidebarConfig.js
```diff
@@ -103,6 +103,22 @@ const config = {
           label: _('Trial Balance'),
           route: '/report/trial-balance',
         },
+        {
+          label: _('GSTR1'),
+          route: '/report/gstr-1',
+          hidden: async () => {
+            const { country } = await frappe.getSingle('AccountingSettings');
+            return country !== 'India';
+          },
```
18alantom:
```javascript
hidden: () => frappe.AccountingSettings.country === 'India',
```

No need to fetch from DB `AccountingSettings` is  cached on the `frappe` object during startup.

src/sidebarConfig.js
```diff
@@ -103,6 +103,22 @@ const config = {
...
+        },
+        {
+          label: _('GSTR2'),
+          route: '/report/gstr-2',
+          hidden: async () => {
+            const { country } = await frappe.getSingle('AccountingSettings');
+            return country !== 'India';
+          },
```
18alantom: Same as the comment above.

reports/GoodsAndServiceTax/GSTR1View.js
```diff
@@ -1,10 +1,11 @@
 const title = 'GSTR 1';
 import baseConfig from './BaseViewConfig';
+import { DateTime } from 'luxon';
```
18alantom: Unnecessary import?

models/doctype/Item/Item.js
```diff
@@ -111,10 +111,17 @@ export default {
         }
       },
     },
+    {
+      fieldname: 'itemCode',
+      label: 'Item Code',
+      fieldtype: 'Int',
+      placeholder: 'Item Code',
```
18alantom: `itemCode` at least in ERPN serves as a unique identifier, this purpose is already served in FB by `name` (_Item Name_).

Any particular reason for adding this in? ~Also if it's required why _Int_? Imo should be alphanumeric, _Data_ seems apt.~

reports/GoodsAndServiceTax/BaseGSTR.js
```diff
@@ -48,7 +48,7 @@ class BaseGSTR {
     row.rate = 0;
     row.inState = true;
     row.reverseCharge = !party.gstin ? 'Y' : 'N';
-    ledgerEntry.taxes.forEach(tax => {
+    ledgerEntry.taxes && ledgerEntry.taxes.forEach(tax => {
```
18alantom:
```javascript
ledgerEntry.taxes?.forEach(...)
```

reports/TrialBalance/viewConfig.js
```diff
@@ -30,16 +30,7 @@ export default {
...
   ],
-  linkFields: [
-    {
-      label: 'Clear Filters',
-      type: 'secondary',
-      action: async report => {
-        await report.getReportData({});
-        report.usedToReRender += 1;
-      }
-    }
```
18alantom, line 41: Any reason for removing the _Clear Filter_ from this and General Ledger?

I noticed, prior to this PR this action wasn't implemented, I don't think we should remove it. If it doesn't work we can fix it.

src/pages/Report.vue
```diff
@@ -2,7 +2,16 @@
...
+      <template slot="actions">        
+        <Button
+          @click="link.action(reportData, filters)"
+          v-for="link of report.linkFields"
+          :key="link.label"
+          type="primary"
+          class="ml-2 text-white text-xs"
+        >
+          {{ link.label }}
+        </Button>
```
18alantom: This will cause all actions to be primary and overpopulate the _title bar_.

Export probably shouldn't be a primary action.

Also Export as JSON should be really be _Export_ drop down menu and _JSON_ as an item cause we'll be adding more filetypes `xlsx`, `pdf` and `csv`.

Items such as _Clear Filters_ can be under a different drop down menu such as _Actions_ or something.

reports/GoodsAndServiceTax/GSTR1.js
```diff
@@ -14,10 +14,10 @@ class GSTR1 extends BaseGSTR {
     const data = await this.getCompleteReport('GSTR-1', filters);
 
     const conditions = {
-      B2B: row => row.gstin,
+      'B2B': row => row.gstin,
       'B2C-Large': row => !row.gstin && !row.inState && row.invAmt >= 250000,
-      'B2C-Small': row =>
-        !row.gstin && (row.inState || (row.inState && row.invAmt < 250000))
+      'B2C-Small': row => !row.gstin && (row.inState || (row.inState && row.invAmt < 250000)),
+      'Nil Rated, Exempted and Non GST supplies': row => (row.rate === 0), // this takes care of both nil rated, exempted goods
```
18alantom, line 21: Can this have a shorter name that doesn't break the UI? 😅

accounting/gst.js
```diff
@@ -0,0 +1,197 @@
...
+  return filePath;
+}
+
+export function promptWhenGstUnavailable() {
+  return new Promise((resolve) => {
+    showMessageDialog({
+      message: _('Export failed!'),
+      description: _(
+        'Report cannot be exported if company gst details are not configured'
+      ),
```
18alantom: The title (`message`) can be "Export Failed", no need for punctuation in the title. Also full stop at the end of `description`.

p.s. I know this is very pedantic but these are almost zero cost improvements to maintain consistency

models/doctype/Item/Item.js
```diff
@@ -111,10 +111,17 @@ export default {
...
+    },
```
18alantom, line 119: Any reason for _Int_ `itemCode`? Also any reason for `itemCode`, in ERPN `item_code` is used as a unique identifier, here `name` (_Item Name_) already serves that purpose.

Imo if this is required we can use _Data_ for an alphanumeric identifier.

reports/GoodsAndServiceTax/BaseViewConfig.js
```diff
@@ -7,76 +9,46 @@ export default {
...
-      fieldname: 'toDate'
-    }
+      fieldname: 'toDate',
+      default: DateTime.local().toISODate(),
+    },
   ],
   linkFields: [
     {
-      label: 'Export',
+      label: 'Export as JSON',
```
18alantom: Imo requires UX refactor, I've mentioned it in some other comment.

reports/GoodsAndServiceTax/BaseViewConfig.js
```diff
@@ -7,76 +9,46 @@ export default {
...
+      fieldname: 'place',
     },
     {
       fieldtype: 'Date',
       label: 'From Date',
       size: 'small',
       placeholder: 'From Date',
-      fieldname: 'fromDate'
+      fieldname: 'fromDate',
+      default: DateTime.local().minus({ months: 3 }).toISODate(),
```
18alantom: This should be a Function, cause `DateTime` will be evaluated on startup, so if someone leaves the software running (i.e. no restart), default will contain stale values.

reports/GoodsAndServiceTax/BaseViewConfig.js
```diff
@@ -7,76 +9,46 @@ export default {
...
     },
     {
       fieldtype: 'Date',
       label: 'To Date',
       size: 'small',
       placeholder: 'To Date',
-      fieldname: 'toDate'
-    }
+      fieldname: 'toDate',
+      default: DateTime.local().toISODate(),
```
18alantom: Same as above

18alantom (changes requested):
@ankitsinghaniyaz I've added some comments, let me know once you're through with them, I'll re-review and we can merge.

I have yet to check report generation and calculations.

models/doctype/Item/Item.js
ankitsinghaniyaz: `itemCode` in the Indian GST context is the HSN/SAC code which should be an integer and should be required. I wanted to add that but missed. Making this field required if GST is configured.

package.json
ankitsinghaniyaz: It was still being imported in `BaseViewConfig.js` and was throwing an error, have removed the import and the packages.

reports/GoodsAndServiceTax/BaseGSTR.js
ankitsinghaniyaz: I somehow don't like the `?` syntax yet, but have updates for consistency.

reports/GoodsAndServiceTax/BaseViewConfig.js
ankitsinghaniyaz: Ok checking the other comment.

reports/GoodsAndServiceTax/GSTR1.js
ankitsinghaniyaz: I think we'll have to keep this and update the UI to accommodate this :sweat_smile.

P.S. We are not exporting this for reports but this is a useful filter

reports/TrialBalance/viewConfig.js
ankitsinghaniyaz: Yeah, we'll have to make it work I guess. I've removed them so that we can add them later as required.

src/pages/Report.vue
ankitsinghaniyaz: This makes sense and we will convert it into a drop-down like an option once we start adding multiple export options.

And I think GST report export JSON action should be a primary action here as that's what it will be used for.

ankitsinghaniyaz: And have removed the hard coding of type primary and it will not take effect from the linkField values.

src/saveReportAsJson.js
ankitsinghaniyaz: Totaly :100: 

Have identified a way to add toast too, but for now opening the directly seems like a good thing to do.

accounting/gst.js
ankitsinghaniyaz: Without the sleep when exporting the invoices the results don't show up sometimes.

reports/TrialBalance/viewConfig.js
18alantom: >  I've removed them so that we can add them later as required.

This doesn't make sense 😅

The problem I see is that someone'll end up implementing a whole other thing to add _Clear Filter_ if we remove it. So since it's already there might as well leave it in. I don't mind pushing commits to fix it.

models/doctype/Item/Item.js
18alantom: Ah okay got it. Might as well add in the conditional required before merge?

Also we should call this **HSN/SAC** to avoid confusion and add it to Item's `RegionalChanges.js` file as opposed to `Item.js`.

Reason for this is to make future refactors to accommodate regional changes less painful.

Check [this](https://github.com/frappe/books/pull/266) for reference. Also the ERPNext implementation, we can supply the codes too.

src/pages/Report.vue
18alantom: Alright, I still find the blue _Export as JSON_ button too garish.

The issue is that the export button will be there (v0.0.5) on other doctypes and reports too, so this will break consistency as it won't be blue everywhere.

Anyways we can leave it in for now, I'll push the changes when taking care of exports.

18alantom (changes requested):
Just two things:
- Item Code to HSN/SAC and shifting it to a regional file.
- Undelete the Clear Filter.

I'll push commits for the other small things, either in this PR or later as fixes.

ankitsinghaniyaz:
Sure @18alantom let me push these changes and ping you.

18alantom (approved):
LGTM now, still haven't cross checked the calculations, will do it when the report is complete, also made a few minor changes:
- Rebased onto `master`
- HSN code is not `required: 1`
- Remove HSN code from `SalesInvoiceItem.js` cause it's in `RegionalChanges.js`
- Renamed *Clear Filter* to *Reset Filters* and moved it to report since it's common and existing implementation is dead code.
- Added group for *Export*, all exports should be under it
- Remove unnecessary Promise wrap around `showMessageDialog`, reworded the message, action calls to receive `getReportData` so that filtered rows are fetched.

PR #299, opened 2022-01-11
Fetching fresh document and avoiding cached one to prevent error

piyushsinghania:
cc: @ankitsinghaniyaz @18alantom 

depends on https://github.com/frappe/frappejs/pull/158


18alantom (approved):

18alantom:
LGTM, the build is breaking for some unrelated reasons ideally, the cached value should be invalidated and updated, but for now merging this.

PR #300, opened 2022-01-11
Updated the layout of checkbox

piyushsinghania:
The checkbox was breaking initially- 
image

_Update the checkbox slightly_

off -
image

checked - 
image

cc: @ankitsinghaniyaz @18alantom

### DIFF
--- a/src/components/Controls/Check.vue
+++ b/src/components/Controls/Check.vue
@@ -74,7 +74,7 @@ export default {
     return {
       offBorderColor: 'rgba(17, 43, 66, 0.201322)',
       offColor: '#FFFFFF',
-      color: '#2490EF'
+      color: '#A1ABB4'
     };
   },
   computed: {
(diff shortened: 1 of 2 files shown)


18alantom (changes requested):
@piyushsinghania I didn't know the checkbox broke on linux. Here are a few screenshots of how it appears on my system:

Screenshot 2022-01-12 at 11 26 49
Screenshot 2022-01-12 at 11 27 32
Screenshot 2022-01-12 at 11 27 38
Screenshot 2022-01-12 at 11 27 55
Screenshot 2022-01-12 at 11 28 01

Can you make it look this way without the UX breaking? Primarily:
- Maintain width and styling with other elements, the report checkbox should have BG and match the width, as of now even I'm not a fan of the background appearance but will change them all together when doing a UI fix later on.
-  Don't use the accent color for the checkbox, it makes the checkbox standout among other elements even when no special importance is given to it. As of now don't change the coloring.

piyushsinghania:
I have made the suggested changes @18alantom 
cc: @ankitsinghaniyaz 

image

image

image

18alantom:
LGTM, failing build is unrelated, hence merging.

PR #310, opened 2022-01-17
Added pesa support to gstr data for proper validation

piyushsinghania:
Fixed the validation issues of GST data export. Added **pesa** support.

Issue - 
image

Expected - 
image

CC: @18alantom @ankitsinghaniyaz

### DIFF
--- a/accounting/gst.js
+++ b/accounting/gst.js
@@ -139,7 +139,7 @@ async function generateB2bData(rows) {
       idt: DateTime.fromFormat(row.invDate, 'yyyy-MM-dd').toFormat(
         'dd-MM-yyyy'
       ),
-      val: row.invAmt,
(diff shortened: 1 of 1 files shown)


18alantom:
@piyushsinghania thanks for bringing this to my notice, but I'll be closing the PR cause of invalid use of `pesa` (I'm almost done with the fix)

Here's why: since _raw_ `knex` queries pull strings from the db, wrapping pesa around a bunch of calculations `pesa(a + b * c)` will lead to very incorrect values in the future if say `a` is a string.

Also doing that nullifies the use of a library such as `pesa`.

piyushsinghania:
If there's a fix than it's great. You have much better idea of Pesa @18alantom 😄

PR #320, opened 2022-01-20
refactor: inline frappejs code into books

18alantom:
Moved relevant code from `frappejs` into `books`. Used [this](https://stackoverflow.com/a/8396318/9681690). There's some stuff included that isn't used in `books`, but I think it may be of use later hence added.

This results in a ~57% reduction in the `frappejs` code.

More context on why: https://github.com/frappe/frappejs/issues/164


taha2002 (commented):
nice

PR #326, opened 2022-01-27
refactor: recode and replace the bar chart

18alantom:
Same as: https://github.com/frappe/books/pull/325 but for the Profit and Loss bar chart.
This also marks the removal of `frappe-charts` from Frappe Books.


18alantom:
_p.s. there is duplicate code between the LIne and the Bar charts, I'll DRY it later when I get time. The primary similarity is the grid and axis code, the rest mostly just looks the same._

PR #330, opened 2022-02-02
fix(ui): temporarily disable maximize

18alantom:
Removing maximize and window resize temporarily.

Rationale for this:
- The dashboard wasn't designed with resizing in mind, hence it doesn't maximize (or resize gracefully).
- This makes the dashboard a bit buggy.
- Not removing it permanently, will add it back in later.


18alantom:
On second thought, won't be merging this. The main issue with chart resizing got fixed on rewrite, so this is not as required.

PR #334, opened 2022-02-09
Setup Translations - Part 1

### LINKED ISSUE #149: Language issues
I chose Brazil in the initial configuration, but it remains in English.
I can't find an option to change the language.
If you don't have a translation into Portuguese, I would like to help translate, just tell me what I have to do.

18alantom:
Todo
- [ ] Add check for translation updates.
- [ ] Update `translation.js` to generate and make use of translation map.
- [ ] Update setup to fetch and set translation file.
- [ ] Add a CONTRIBUTIONS.md somewhere
- [ ] Add a some translation files
- [x] Add script to generate translation file `.csv`.
- [x] Convert all translation instances to translation template strings.

~Will add more details soon.~

~As of now converting all non-idiomatic occurrences of `t`, eg: `t('value: {}', 0)` to idiomatic ones i.e:~
```javascript
t`value ${0}`
```
~I've reached a bit of an impasse as it seems Vue 2.x doesn't support this syntax in SFC templates 😐~ 
Impasse, passed, Vue is now at 3 😎

As for why am I doing this? Basically reducing the amount of regex I have to write, more regex == more misery.


taha2002 (approved):

18alantom:
This PR basically sets the stage for translations to take place, the next one will enable it.
Due to how it will be handled I need to push it through two separate PRs.

PR #341, opened 2022-02-16
Setup Translations - Part 2 Use Translations

### LINKED ISSUE #149: Language issues

18alantom:
Todo
- [x] Check builds
- [x] Update Setup Wizard to fetch and set translation file.
- [x] Add a [Contribution Notes](https://github.com/frappe/books/wiki/Contributing-Translations)
- [x] Add check for translation updates.
- [x] Update `translation.js` to generate and make use of translation map.
- [x] Add a some translation files
- [x] Add script to generate translation file `.csv`.
- [x] Convert all translation instances to translation template strings.

this will close #149

### DIFF
--- a/frappe/index.js
+++ b/frappe/index.js
@@ -7,6 +7,8 @@ const {
   DEFAULT_DISPLAY_PRECISION,
 } = require('./utils/consts');
 const { markRaw } = require('vue');
+const { ipcRenderer } = require('electron');
+const { IPC_ACTIONS } = require('@/messages');
 
 module.exports = {
   initializeAndRegister(customModels = {}, force = false) {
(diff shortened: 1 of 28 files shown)


18alantom:
There's a strange issue where reloading the window on language change causes the Production version to hang around 40% of the time.

I think this is cause of the  `mainProcess` `fetch`, which should be occurring before the reload.

Nevertheless will try and fix this issue later, translations are working properly, only strings need to be added.

Issue #176, opened 2020-10-01
Profit and Loss report erroneous


bynicolas:
The **Profit and Loss** report is erroneous when you select the `Yearly` view.

it seems to be an issue with the date range that don't match the expected fiscal year dates set in the *Settings*.

For instance, I expect to get totals for each accounts in the range `2020-01-01` to `2020-12-31`, but some values are missing from the total because the reports thinks its in an other fiscal year.

Anything before 2020-04-01 is considered 2019
Anything up to 2021-03-31 is considered 2020
And neither match the expected fiscal year dates

![image](https://user-images.githubusercontent.com/10983565/94771280-e2c41600-0384-11eb-9547-c117a02ac768.png)
![image](https://user-images.githubusercontent.com/10983565/94771251-d3dd6380-0384-11eb-8d92-b1af6e8c6d73.png)

18alantom:
Seems to be caused by the _date off by one day bug_, this may have been fixed in #220. Not closing cause it needs further evaluation.

18alantom:
Fixed in #305

bynicolas:
Is this really fixed? I see the same behaviour in v0.4.2-beta.0.

A yearly P/L report from `2021-01-01` to `2021-12-31` would give a report range from `2020-2021`

A half yearly report for the same date range would give from `Oct 2020 - Mar 2021` and `Apr 2021 - Sep 2021`

Monthly and Quarterly P/L reports seems to fit the requested range

18alantom:
Will close this when it's fixed and tests have been added to verify it.

18alantom:
<img width="1004" alt="Screen Shot 2022-05-27 at 01 13 04" src="https://user-images.githubusercontent.com/29507195/170565078-673a59b1-589e-4d05-b8e9-e67220b11a05.png">

All reports have been rewritten, PnL report now shows the **Until Date** of the interval (monthly, quarterly, etc), will add a description tooltip soon to make things clearer.

PR #375, opened 2022-04-01
fix: #176 erroneous Profit ans Loss reports

4silvertooth:
Hello,

This commit is an attempt to fix erroneous Profit ans Loss reports, current reports are coded assuming that fiscal year will always or mostly start from April and not other dates 
https://github.com/frappe/books/blob/4c706afe17e40c93861a0f2d40e7f4b4b0f384f1/reports/FinancialStatements/FinancialStatements.js#L194
here in  `Half Yearly` if `month > 3` (assumes fiscal year either starts from Apr) would give proper results only if fiscal starts from Apr but not otherwise and would fall back to `Oct ${year - 1} - Mar ${year}`. This commit takes into consideration the fiscal data to fix that.

Also DateTime library `luxon` just gives out quarter assuming first quarter starts from January, 
https://github.com/frappe/books/blob/4c706afe17e40c93861a0f2d40e7f4b4b0f384f1/reports/FinancialStatements/FinancialStatements.js#L187
although Quarters reports seems to be right but the underlying tabulation doesn't seem to be correct if the labels are changed to Q1..Q4 it gives erroneous results, for example if a Company with fiscal year starting from January would give Q1 result in first column but any company who's fiscal starts other than January the first column starts with Q2 or Q3. Just change https://github.com/frappe/books/blob/4c706afe17e40c93861a0f2d40e7f4b4b0f384f1/reports/FinancialStatements/FinancialStatements.js#L187 to      
```
return {
        1: `Q1 Jan ${year} - Mar ${year}`,
        2: `Q2 Apr ${year} - Jun ${year}`,
        3: `Q3 Jun ${year} - Sep ${year}`,
        4: `Q4 Oct ${year} - Dec ${year}`,
      }[quarter];
    }
``` 
for the actual result. 
pl2

This commit also fixes that but the column label would now will be labeled Q1..Q4 with proper years.
pl3

The current ProfitLoss report's filter `start date` and `end date` input 
behaves as though changing it changes the `fiscal start and fiscal end` and not being just a period of data which we want to look through. If the FiscalYearStart/End month is from April to March and if the filter is set to `Jan to Dec` the reports should display data for `Jan-Mar FY{-1}` and `Apr-Dec FY` but it displays reports as though the Fiscal `start` and `end` has changed, this commit also takes care of that, with showing data window from filters but underlying report stick to fiscal year data. 

**More testing/reviewing required.**

### DIFF
--- a/reports/Cashflow/Cashflow.js
+++ b/reports/Cashflow/Cashflow.js
@@ -1,6 +1,6 @@
 import frappe from 'frappe';
 import { DateTime } from 'luxon';
-import { getPeriodList } from '../FinancialStatements/FinancialStatements';
+import { getPeriodList, getFiscalYear } from '../FinancialStatements/FinancialStatements';
 
 class Cashflow {
   async run({ fromDate, toDate, periodicity }) {
@@ -24,7 +24,8 @@ class Cashflow {
       .whereBetween('date', [fromDate, toDate])
       .groupBy(dateAsMonthYear);
 
-    let periodList = getPeriodList(fromDate, toDate, periodicity);
+    let fiscalYear = await getFiscalYear();
+    let periodList = getPeriodList(fromDate, toDate, periodicity, fiscalYear);
 
     let data = periodList.map((periodKey) => {
       let monthYear = this.getMonthYear(periodKey, 'MMM yyyy');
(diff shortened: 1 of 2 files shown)


18alantom (approved):
@4silvertooth awesome PR description. I think this should do for now. Will add tests for the reports soon.

PR #477, opened 2022-10-08
feat: Add Simplified Chinese translation

wcxu21:
Add Simplified Chinese translation

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,17 +128,18 @@ If you want to contribute code then you can fork this repo, make changes and rai
 
 ## Translation Contributors
 
-| Language   | Contributors                                                                       |
-| ---------- | ---------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


18alantom:
Hey @wcxu21, can you use UTF-8 encoding, the characters don't render properly.

wcxu21:
Hi
I use the Office 2021 version to edit CSV files, and by default I don't need to adjust the encoding settings. 
 I myself upload the translation file after it is normal, no garbled situation. However, when the code is submitted, the preview will show garbled characters. I don't know if it was caused by a Github preview problem. 
You can see in the source file I submitted that it is displayed correctly.

https://github.com/wcxu21/books/blob/master/translations/zh-CN.csv

18alantom:
The submitted source had the correct characters, I've used [that](https://github.com/wcxu21/books/blob/master/translations/zh-CN.csv) instead.

PR #521, opened 2023-01-07
Rtl support

zaqoutabed:
Add RTL Support by using tailwindcss-rtl package.

Screenshot from 2023-01-07 13-58-21


18alantom:
Looks good for the most part, with just a few minor issues that occur mostly due to hiding the sidebar:
1. The reveal sidebar button (bottom left red box) should be on the right.
2. The macOS traffic lights (top left red box) should be either on the right or there should be some padding to prevent the overlay.
3. The right side padding (top right red box) used for the traffic lights should be removed unless the traffic lights are moved to the other side.

Screen Shot 2023-01-13 at 11 20 15


I'm merging this PR. The fix for this can be raised through another PR.

PR #528, opened 2023-01-16
translation: edit/fix few translation in gujarati that came to notice

4silvertooth:
A few remaining fix to Gujarati translation, also the last time I `yarn script:translate -l gu` it removed all the quotes marks `"` around strings but other language translations has them so this update and fix includes the quotes around strings.


18alantom:
@4silvertooth the double quotes are not strictly required. The CSV standard ([RFC-4180](https://www.rfc-editor.org/rfc/rfc4180#section-2)) requires it only if an item in a row contains a double quote or a comma:

>  6.  Fields containing line breaks (CRLF), double quotes, and commas should be enclosed in double-quotes.

I know the other languages have it, that's cause the contributors created the files that way, but since the `gu` translations didn't have it initially, it's better to not add double quotes unless strictly required (for example [this](https://github.com/frappe/books/pull/528/files#diff-06877399e7a28c43e0ec80b10478b3a621bf8141a91bb30c7b5772f2def27967L541) line) as it will be removed when the script is re-run to add new lines.

Could you re-run the script and push a commit, it should preserve the translation fixes. Other than that LGTM.

4silvertooth:
> it's better to not add double quotes unless strictly required

The script adds the quotes even if the csv doesn't have quotes, any future fix to the gu translation will require the quotes to be removed manually.

@18alantom No idea why the checks are failing.

18alantom:
@4silvertooth my bad, you're right the script adds double quotes even if there are spaces. 

Failing checks are unrelated to the PR. It's caused by a dependency mismatch after one of the packages required by the workflow got updated.

LGTM, merging it.

PR #535, opened 2023-01-28
feat: batch wise stock balance report

akshayitzme:
- [x] Batch Wise Stock Balance Report
image

### DIFF
--- a/backend/database/bespoke.ts
+++ b/backend/database/bespoke.ts
@@ -136,9 +136,9 @@ export class BespokeQueries {
     db: DatabaseCore,
     item: string
   ): Promise<boolean> {
-  /*
-    * to check if an item of given name is present in the SLE
-  */
+    /*
+     * to check if an item of given name is present in the SLE
+     */
     const query = db.knex!(ModelNameEnum.StockLedgerEntry)
       .select('item')
       .where('item', item);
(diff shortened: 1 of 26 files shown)


18alantom:
@akshayitzme why is this a separate report? Shouldn't Batch be a filter in the Stock Balance report itself?

rtdany10:
> @akshayitzme why is this a separate report? Shouldn't Batch be a filter in the Stock Balance report itself?

Stock Balance shows total available stock of an item in a particular warehouse.
This report should batch wise available qty of the particular item in a particular warehouse.

Batch wise report:
Item A - Batch 1 - Warehouse X - 5 Nos
Item A - Batch 2 - Warehouse X - 5 Nos
Item A - Batch 3 - Warehouse Y - 5 Nos
Item A - Batch 4 - Warehouse Z - 5 Nos

Stock Balance report:
Item A - Warehouse X - 10 Nos
Item A - Warehouse Y - 5 Nos
Item A - Warehouse Z - 5 Nos

18alantom:
@rtdany10 I understand your point.

But taking into consideration simplicity, I still hold that this report should be just a filter in the Stock Balance report.

---

Here's a way to do it, consider the following filters:
1. Show Batch-wise
2. Batch No

If either of them is set then an additional Batch column is displayed and the inventory quantities are shown split by batch thereby replicating the functionality of this report.

---

Conceptually a unique combination of _Item, Batch, and Location_ is just a compound key to a quantity variable (I haven't checked the Batch No. PR, I hope it's been handled as such). The _Show `$KEY_ITEM`-wise_ filter just drops an item from the key and aggregates the quantities.

In fact, the base Stock Balance report can be further expanded to show aggregate quantities across all Locations when a _Show Location-wise_ flag is unset. This feature wouldn't entail a separate Location-wise Stock Balance report.

PR #548, opened 2023-02-13
feat: UOM Conversion Factor  

akshayitzme:
- [x]  UOM Conversion Factor Schema
image

- [x] Enable UOM Conversion Factor
image

### DIFF
--- a/models/baseModels/InvoiceItem/InvoiceItem.ts
+++ b/models/baseModels/InvoiceItem/InvoiceItem.ts
@@ -1,4 +1,4 @@
-import { Fyo } from 'fyo';
+import { Fyo, t } from 'fyo';
 import { DocValue, DocValueMap } from 'fyo/core/types';
 import { Doc } from 'fyo/model/doc';
 import {
@@ -13,6 +13,7 @@ import { ValidationError } from 'fyo/utils/errors';
 import { ModelNameEnum } from 'models/types';
 import { Money } from 'pesa';
 import { FieldTypeEnum, Schema } from 'schemas/types';
+import { safeParseFloat } from 'utils/index';
 import { Invoice } from '../Invoice/Invoice';
 import { Item } from '../Item/Item';
 
@@ -23,6 +24,10 @@ export abstract class InvoiceItem extends Doc {
   parentdoc?: Invoice;
   rate?: Money;
   quantity?: number;
+  transferQty?: number;
+  stockUOM?: string;
+  uom?: string;
+  UOMConversionFactor?: number;
   tax?: string;
   stockNotTransferred?: number;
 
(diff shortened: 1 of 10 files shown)


schemas/app/UOMConversionFactor.json
```diff
@@ -0,0 +1,24 @@
...
+    {
+      "fieldname": "uom",
+      "label": "UOM",
+      "fieldtype": "Link",
+      "target": "UOM",
+      "create": true,
+      "required": true
+    },
+    {
+      "fieldname": "value",
```
18alantom, line 15: This should be called `conversionFactor` or something else that indicates its purpose, not `value`

schemas/app/inventory/StockMovementItem.json
```diff
@@ -26,13 +26,37 @@
...
       "fieldname": "quantity",
       "label": "Quantity",
       "fieldtype": "Float",
       "required": true,
       "default": 1
     },
+    {
+      "fieldname": "stockUOM",
+      "label": "Stock UOM",
+      "fieldtype": "Data",
```
18alantom, line 45: Why is this not a Link field?

models/baseModels/InvoiceItem/InvoiceItem.ts
```diff
@@ -23,6 +24,10 @@ export abstract class InvoiceItem extends Doc {
   parentdoc?: Invoice;
   rate?: Money;
   quantity?: number;
+  transferQty?: number;
+  stockUOM?: string;
+  uom?: string;
+  UOMConversionFactor?: number;
```
18alantom, line 30: Should not start with caps, it should be `uomConversionFactor`.

18alantom (changes requested):
@akshayitzme
1. Why is `stockUOM` not a link field everywhere?
2. `UOMConversionFactor` (the table in Items) should be displayed only if inventory is enabled
3. The StockMovement table and StockTransfer tables should only show the unit the user has selected for transfer (eg Box), conversion to the stock unit (eg Unit) should take place behind the scenes. This is unacceptable:
Screen Shot 2023-02-17 at 17 32 30 I'll be creating a General Form view for larger entries so this additional info can be displayed in the quick edit like how Invoice Items information is displayed.

4. The field naming and labeling seems to be all over the place 😅 Screen Shot 2023-02-17 at 17 32 30
- There seems to be a bug, in the above image if the user-selected unit is Box and conversion rate is 10 shouldn't the quantity as per stock UOM i.e. Unit be 10 and not the other way round?
- Why is the label `"UOMs"`
- The variable naming `quantity`, `transferQty`, `stockUOM`, `uom`, `UOMConversionFactor` and `unit` (on Item) seems very confusing, it's not apparent which of these values is for user convenience and which one is used for maintaining the Stock Balance.

18alantom:
Closing this cause of #557

PR #580, opened 2023-03-17
always show Period Selector

zaqoutabed:
Fix: if there is no data i need to be able to change the period, and show new button if there is no data


18alantom:
Why do you need to see the period selector when there's no data if it doesn't do anything?

Closing cause use case is not understood and cause it causes bad UI:

Screen Shot 2023-03-20 at 08 44 26

zaqoutabed:
> @18alantom 

when i change the period to `This Month` for example
and there is no data the selector will disappear and I cannot select another period

18alantom:
@zaqoutabed got it, I've fixed it in this https://github.com/frappe/books/pull/581

PR #599, opened 2023-04-20
feat: informative links

18alantom:
- fix(ui): hanging border when using bg header - focus not found issue - save shortcut in Settings
- chore: type Popover.vue
- refactor: simplify autocomplete


18alantom:
Closing cause I've lost context on this. Will redo later.

PR #657, opened 2023-06-07
fix: Price List PR #653

18alantom:
Simplifies and Fixes #653 

Changes:
- Fixed invoice item price fetch logic
- ItemPrice renamed to PriceListItem
- Removed the following `PriceListItem` fields `enabled`, `buying`, `selling`, `party`, `batch`, `validFrom`, `validUpto`.
- Removed the following from `PriceList` fields `isUomDependent`.
- Renamed `enabled`, `buying,` `selling`


schemas/app/PriceList.json
```diff
@@ -10,36 +10,28 @@
...
-      "default": false,
-      "required": true
-    },
-    {
-      "fieldname": "selling",
-      "label": "Selling",
-      "fieldtype": "Check",
-      "default": false,
-      "required": true
+      "default": true
```
akshayitzme, line 22: why `isSales` default value set to `true` ?

schemas/app/PriceListItem.json
```diff
@@ -0,0 +1,31 @@
...
+    {
+      "fieldname": "rate",
+      "label": "Rate",
+      "fieldtype": "Currency",
+      "required": true,
+      "section": "Item"
+    }
+  ],
+  "quickEditFields": ["item", "unit", "rate"]
+}
```
akshayitzme, line 31: `PriceListItem` child table is invisible cause `tableFields` are not defined here !.

image

schemas/app/PriceList.json
18alantom: Seems like a better default. Users make Sales more often than Purchases.

akshayitzme: okay got it

schemas/app/PriceListItem.json
18alantom: Thanks! FB _really_ needs UI tests 🙈

PR #698, opened 2023-07-16
set all visible types in ”@types” packages

zaqoutabed:


18alantom:
@zaqoutabed I didn't understand what this PR is for, could you explain?


Linting breaks cause of this. This line:
```typescript
import installExtension, { VUEJS3_DEVTOOLS } from 'electron-devtools-installer';
```
causes it because `VUEJS3_DEVTOOLS` is not exported and `VUEJS_DEVTOOLS` is a different dev tool, so switching won't help.

zaqoutabed:
I think I am wrong sorry for that I didn't notice that VUEJS DEVTOOLS has different type. I will close this PR.
thank you for explanation.

PR #769, opened 2023-11-02
Add regional schema for Switzerland

### LINKED ISSUE #768: Tax ID (VAT) not available for Switzerland
Trying to use Frappe Books to do simple accounting for a Swiss small business. I need to specify the VAT number on invoices, but Frappe books doesn't allow be to set it.

digitalentity:
Fixes  #768, allowes to specify Tax ID (VAT number) if country is `Switzerland`, propagates the `taxId` attribute to print object to be usable in the print templates.

### DIFF
--- a/schemas/regional/ch/AccountingSettings.json
+++ b/schemas/regional/ch/AccountingSettings.json
@@ -1,0 +1,21 @@
+{
+  "name": "AccountingSettings",
+  "fields": [
+    {
+      "fieldname": "taxId",
+      "label": "Tax ID",
+      "fieldtype": "Data",
+      "placeholder": "CHE-123.456.789",
+      "section": "Default"
+    }
+  ],
+  "quickEditFields": [
+    "fullname",
+    "email",
+    "companyName",
+    "country",
+    "fiscalYearStart",
+    "fiscalYearEnd",
+    "taxId"
+  ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ yarn-error.log*
 *.njsproj
 *.sln
 *.sw?
+*.code-workspace
 
 #Electron-builder output
 /dist_electron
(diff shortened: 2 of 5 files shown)


schemas/regional/ch/AccountingSettings.json
```diff
@@ -0,0 +1,22 @@
...
+  ],
+  "quickEditFields": [
+    "fullname",
+    "email",
+    "companyName",
+    "country",
+    "currency",
+    "fiscalYearStart",
+    "fiscalYearEnd",
+    "taxId"
```
mildred: I believe this list should be compared to

https://github.com/frappe/books/blob/8bf0a5f7362fbdcecf7d036061a3387adf889bb5/schemas/app/AccountingSettings.json#L125-L132

I suppose the field `currency` was copied from the Indian regional schema. Does it makes sense for Swizerland?

mildred (commented):
I believe this might be useful for more than just Swizerland in the end. Probaby all EU countries

schemas/regional/ch/AccountingSettings.json
digitalentity: While CHF is the national currency in Switzerland it is not uncommon for accounting to be done in EUR, especially if this is a company selling services in EU.

digitalentity:
EU is a bit more complicated, but EU countries may benefit from a shared regional schema, but that schema will likely be different than Swiss (VAT numbers are not the same format).

mildred:
Can you run the linter over the sources files, and perhaps remove the `currency` quick edit field?

Files needing linting:

```
[warn] schemas/regional/ch/AccountingSettings.json
[warn] schemas/regional/index.ts
```

command to run:

    npx prettier -w schemas

digitalentity:
Done, also squashed the commits

schemas/regional/ch/AccountingSettings.json
mildred: If this is intentional, no problem, I supposed something like this. But if this was just copy and paste from Indian regional schema, perhaps this did not need to be included.

PR #778, opened 2023-11-29
add the option to turn off item account filtering

jkroso:
I'm not sure this is good accounting practice but I prefer to be able to record asset purchases the same way I record expenses.

### DIFF
--- a/models/baseModels/Item/Item.ts
+++ b/models/baseModels/Item/Item.ts
@@ -58,16 +58,28 @@ export class Item extends Doc {
   };
 
   static filters: FiltersMap = {
-    incomeAccount: () => ({
-      isGroup: false,
-      rootType: AccountRootTypeEnum.Income,
-    }),
-    expenseAccount: (doc) => ({
-      isGroup: false,
-      rootType: doc.trackItem
-        ? AccountRootTypeEnum.Liability
-        : AccountRootTypeEnum.Expense,
-    }),
+    incomeAccount: (doc) => {
+      if (doc.fyo.singles.AccountingSettings?.itemAccountFiltering) {
+        return {
+          isGroup: false,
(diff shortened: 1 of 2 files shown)


mildred:
Basically, you're adding a toggle to allow selecting any kind of account for an item. Would it be possible to have a behavior that would work for you without introducing a new toggle? What kind of account are you using that is today filtered out?

@18alantom what do you think?

jkroso:
Well I'm trying to assign an item to my "tools" account which is a fixed asset account. So yeah I could just change the default filtering instead if you think that would work for everyone.

mildred:
Perhaps a better option would be to keep the account filtering but let the user enter accounts outside of the filter if he really wants to.

jkroso:
So add a button in the drop down to turn off filtering in a specific case rather than globally?

18alantom:
@mildred I think since this is a change that can lead to undefined behavior (causing issues by confused users later on) it should be restricted in someway. Perhaps behind a feature flag.

The feature itself could be an option in the drop down that allows turning of filtering (such as what @jkroso suggested).

If not feature flag, the option could be made visible if the list is empty for a given query.

mildred:
@18alantom so, do you think the global option introduced in this PR is the good way to do it or should it specifically be in the feature flag section of the options?

18alantom:
@mildred I think, ideally the feature flag should be to enable _optionally removing filters when entering Link field values_ (i.e. across all Link fields).

The feature flag in this PR completely switches it off for two fields. Disabling filters for a Link field seems like an infrequent requirement.

Also, there have been instances wherein users have needed filters across other fields to be removed. Introducing individual feature flags for each would unnecessarily populate the feature flag section.

mildred:
Implemented in #1022

PR #801, opened 2024-01-14
Electron and Dependencies update

Isaac-GC:
This is a supplemental PR (meant to be an update and replacement for [PR#751](https://github.com/frappe/books/pull/751)).

Updates
`electron` -> 18.3.7 -> 22.3.27
`electron-builder` 24.4.0 -> 24.9.1
`electron-updater` 5.2.1 -> 6.1.7
`better-sqlite3` 7.5.3 -> 9.2.2

`electron-rebuild` -> updated to `@electron/rebuild` w/version 3.4.1

Currently, with updated electron versions, there is an issue where the `node-abi` version can't be automatically detected w/`electron-builder`. To get around this issue, the `node-abi` version was added in the "resolutions" portion of the package.json

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v2
         with:
-          node-version: '16.20.2'
+          node-version: '18.19.0'
 
       - name: Set yarn version
         run: yarn set version 1.22.18
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v2
         with:
-          node-version: '16.20.2'
+          node-version: '18.19.0'
 
       - name: Set yarn version
         run: yarn set version 1.22.18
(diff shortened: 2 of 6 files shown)


18alantom:
Seems like it's not finding pre-built binaries for `better-sqlite3` under Node 16.14.0 and Darwin causing all tests to fail:
https://github.com/frappe/books/actions/runs/7521667651/job/20472776937?pr=801#step:5:17

(`build.yml` runs on macOS since it's the only one that builds for all platforms)

Isaac-GC:
Updated the nodejs version from 16 to 18 in the workflow files and that seemed to fix it.

18alantom:
Checked it locally (ARM macOS 13.5) nothing overtly broken, looks good.

mildred (approved):
Is there a reason why package-lock.json was never committed

Else, LGTM

Isaac-GC:
No, looks like it was removed from the master branch since 2018), but did add in `yarn.lock`

mildred:
I was thinking there might be a legitimate reason why yarn.lock was not included

Isaac-GC:
Since Yarn was implemented as the tool of choice by using the `yarn` command to install dependencies, I understand the `yarn.lock` takes the place of `package-lock.json`. I don't really see a conflict that might happen with having the yarn.lock file (though having both of types of lockfiles present in the repo might cause issues)

@18alantom Is/Was there issues previously with having the yarn.lock file present in the repo and/or should there be a package-lock.json as well?

PR #807, opened 2024-01-16
Added issue templates to help with workflows

Isaac-GC:
Added issue templates to help creation of issues and aid in automated workflows in project tab

### DIFF
--- a/.github/ISSUE_TEMPLATE/2-feature_template.yml
+++ b/.github/ISSUE_TEMPLATE/2-feature_template.yml
@@ -14,6 +14,15 @@ body:
       required: true
 
   - type: textarea
+    id: benefits
+    attributes:
+      label: 'What problem are you trying to solve?'
+      description: Tell us what is the thing you are doing and why this feature would help you in that
+      placeholder: Describe the problem or issue that the feature would solve
+    validations:
+      required: true
+
+  - type: textarea
     id: basic_example
     attributes:
       label: 'Basic Example'
--- a/.github/ISSUE_TEMPLATE/2-feature_template.yml
+++ b/.github/ISSUE_TEMPLATE/2-feature_template.yml
@@ -1,22 +1,22 @@
-name: "💡 Feature Request"
+name: '💡 Feature Request'
 description: Create a new ticket for a new feature request
-title: "💡 [Feature Request] - <title>"
-labels: [ "enhancement" ]
+title: '💡 [Feature Request] - <title>'
+labels: ['enhancement']
 
 body:
   - type: textarea
     id: summary
     attributes:
-      label: "Summary"
+      label: 'Summary'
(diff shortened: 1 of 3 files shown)


Isaac-GC:
@18alantom and @mildred Just tagging you in this PR to see if there other templates you'd like to see or if there are other items in the templates themselves you think we should have.

Thanks :)

18alantom:
The templates look good, should raise the bar for reported issues! I think these three should be fine for now.

A minor worry is fewer issues would be raised due to the required steps (albeit they'd be more meaningful).

Also, clicking on **Report Issue** in Books redirects to the new issue page with a Template: https://github.com/frappe/books/blob/master/src/errorHandling.ts#L195 This might need updating.

.github/ISSUE_TEMPLATE/3-general_question_template.yml
```diff
@@ -0,0 +1,14 @@
...
+
+body:
+  - type: textarea
+    id: summary
+    attributes:
+      label: 'Summary'
+      description: General Question(s) (for Bugs and Feature Requests, please use the appropriate template)
+      placeholder: '...'
+    validations:
+      required: true
```
mildred, line 14: Perhaps we could make use of the discussions GitHub feature for this.

.github/ISSUE_TEMPLATE/2-feature_template.yml
```diff
@@ -0,0 +1,41 @@
...
+
+body:
+  - type: textarea
+    id: summary
+    attributes:
+      label: 'Summary'
+      description: Provide a brief explanation of the feature
+      placeholder: Describe in a few lines your feature request
+    validations:
+      required: true
```
mildred, line 14: I would add also a question :

**What problem are you trying to solve?**
*Tell us what is the thing you are doing and why this feature would help you in that*

mildred (approved):

.github/ISSUE_TEMPLATE/3-general_question_template.yml
Isaac-GC: I figure instead of completely pushing people to github discussions or telegram, we would instead continue giving the ability for questions and then sometime in the future, just get rid of it all together (trying to prevent too much change at once)

Isaac-GC:
> The templates look good, should raise the bar for reported issues! I think these three should be fine for now.
> 
> A minor worry is fewer issues would be raised due to the required steps (albeit they'd be more meaningful).
> 
> Also, clicking on **Report Issue** in Books redirects to the new issue page with a Template: https://github.com/frappe/books/blob/master/src/errorHandling.ts#L195 This might need updating.

Looks like I'll have to update the errorHandling.ts file separately. I can't update it until we have the issue template published due to the urll being different.

PR #827, opened 2024-01-30
Fixing printToPDF from electron window

### LINKED ISSUE #824: Invoice print seems broken after the recent update
The print works fine till you save the invoice in Frappe Books, and it looks intact in the system. But as you save it as PDF it seem broken this is how it looks after save to pdf

image

**Version**: `0.21.0`
**Platform**: `darwin`
**Path**: `/print/SalesInvoice/SINV-23241009`
**Language**: `-`
**Country**: `in`

### LINKED ISSUE #833: generated pdf's are blank
<h2>Description</h2>
while generating the pdf's for invoice they are blank

<h2>Steps to Reproduce</h2>
1. click on the printer icon
2. choose save as PDF
3. generated pdf's are blank

<h2>Info</h2>

**Version**: `0.21.0`
**Platform**: `darwin`
**Path**: `/print/SalesInvoice/SINV-1061`
**Language**: `-`
**Country**: `in`

Isaac-GC:
Fixing an issue with Printing to a PDF.

Hi @mildred @18alantom 

Testing this fix on local devices (linux, windows, and mac arm) it looks fine, but requesting verification JIC.

Thanks!

### DIFF
--- a/main/saveHtmlAsPdf.ts
+++ b/main/saveHtmlAsPdf.ts
@@ -20,14 +20,12 @@ export async function saveHtmlAsPdf(
 
   const printWindow = await getInitializedPrintWindow(htmlPath, width, height);
   const printOptions = {
-    marginsType: 1, // no margin
+    margins: { top: 0, bottom: 0, left: 0, right: 0 }, // equivalent to previous 'marginType: 1'
     pageSize: {
-      height: height * 10_000, // micrometers
-      width: width * 10_000, // micrometers
+      height: height / 2.54, // Convert from centimeters to inches
+      width: width / 2.54, // Convert from centimeters to inches
     },
     printBackground: true,
-    printBackgrounds: true,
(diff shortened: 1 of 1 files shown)


mildred (approved):
It seems the PDF exports page size are off in v0.21.0, do you think that is from a dependency update?

Most probably, I could not save a PDF with your branch until I updated the dependencies... and it fixes the page size issue

Isaac-GC:
> It seems the PDF exports page size are off in v0.21.0, do you think that is from a dependency update?
> 
> Most probably, I could not save a PDF with your branch until I updated the dependencies... and it fixes the page size issue

Correct --> It partly has to due to the dependency updates and specifically, the parameter option changes in the `window.printToPDF`.  While the pdf's generated DO actually contain the expected contents... the actual contents are incredibly tiny due to the changes... But _are_ fixed

PR #831, opened 2024-01-31
Bug fix recorddate inconsistency

Isaac-GC:
After doing some testing in various timezones it looks like this (very ugly) patch successfully fixes the various timestamp inconsistency issues, which resulted in incorrect ledgers, profit/loss, etc

What was happening is that depending on one's timezone, the Date that was saved in the JournalEntry/Ledger would convert the Datetime to a timestamp of `00:00.000` for the local timezone and *then* calculates the GMT timestamp based on the local timezone. This creates an instance where what is saved may not properly reflect what was entered and expected in the related invoices (or other items). 

Here are two test cases showing the issue of what was happening:
---

1. Test case 1 (Myanmar timezone -> `GMT +6.5`)
image

The time is converted to GMT which is 6.5 hours behind Myanmar at that point of time, resulting in saving the datetime as previous date (in this test case, `2022-12-31T17:30.000Z`)  

---

2. Test case 2 (US Pacific timezone -> `GMT -8`)
Screenshot 2024-01-30 193727  

In this case, when the timestamp for the local timezone is `00:00.000`,  GMT time will be 8 hours ahead (`2023-01-01T08:00.000Z`)

### DIFF
--- a/backend/patches/v0_21_0/fixLedgerDateTime.ts
+++ b/backend/patches/v0_21_0/fixLedgerDateTime.ts
@@ -1,0 +1,34 @@
+import { DatabaseManager } from '../../database/manager';
+
+/* eslint-disable */
+async function execute(dm: DatabaseManager) {
+    await dm.db!.knex!('AccountingLedgerEntry')
+        .select('name', 'date')
+        .then((trx: Array<{name: string; date: Date;}> ) => {
+            trx.forEach(async entry => {
+                const entryDate = new Date(entry['date']);
+                const timeZoneOffset = entryDate.getTimezoneOffset();
+                const offsetMinutes = timeZoneOffset % 60;
+                const offsetHours = (timeZoneOffset - offsetMinutes) / 60;
+
+                let daysToAdd = 0; // If behind or at GMT/Zulu time, don't need to add a day
+                if (timeZoneOffset < 0) {
+                    // If ahead of GMT/Zulu time, need to advance a day forward first
+                    daysToAdd = 1; 
+                }
+
+                entryDate.setDate(entryDate.getDate() + daysToAdd);
+                entryDate.setHours(0 - offsetHours);
+                entryDate.setMinutes(0 - offsetMinutes);
+                entryDate.setSeconds(0);
+                entryDate.setMilliseconds(0);
+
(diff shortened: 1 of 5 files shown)


models/Transactional/LedgerPosting.ts
```diff
@@ -90,12 +90,30 @@ export class LedgerPosting {
       return map[account];
     }
 
+    // Timezone inconsistency fix (very ugly code for now)
+    const entryDateTime = this.refDoc.date as string | Date;
+    let dateTimeValue: Date;
+    if (typeof entryDateTime === 'string' || entryDateTime instanceof String) {
```
mildred, line 96: Isn't that equivalent?

mildred (approved):
Good catch. perhaps this could be in a separate function so its scope, input and output  is more clearly understood.

models/Transactional/LedgerPosting.ts
Isaac-GC: It *should be*, yet as we saw with this DateTime fun, there are some weird errors that pop up. By forcing a check I want to make sure that it *is* as expected as I really don't want to revisit this error later.

Isaac-GC:
The plan is to break it out in the future, but as you mentioned we should check the scope of the entirety of the issue. I included it in the function to ensure its scope is only applied to that section and children that use it. 

For the scope of the issue, if you look at the db patch, it shows all of the models affected

backend/patches/v0_21_0/fixLedgerDateTime.ts
```diff
@@ -0,0 +1,34 @@
...
+
+                let daysToAdd = 0; // If behind or at GMT/Zulu time, don't need to add a day
+                if (timeZoneOffset < 0) {
+                    // If ahead of GMT/Zulu time, need to advance a day forward first
+                    daysToAdd = 1; 
+                }
+
+                entryDate.setDate(entryDate.getDate() + daysToAdd);
+                entryDate.setHours(0 - offsetHours);
+                entryDate.setMinutes(0 - offsetMinutes);
```
Isaac-GC, line 22: Using the offsetMinutes and offsetHours to set the values aren't necessary but are there to mirror with the code fix in LedgerPosting.ts

PR #837, opened 2024-02-02
[Bug] - Fix Type error in Colors.vue without default colors

shahzadbinshahjahan:
### Issue

When adding a color field by customizing the form without specifying default values, `Color.vue` throws a Type error because it tries to find the color from an undefined colors array

### Changes Made

Added a condition before the `find` call to return the value if there are no items in the colors array.
Screenshot from 2024-02-02 18 23 29
Screenshot from 2024-02-02 18 21 44

### DIFF
--- a/src/components/Controls/Color.vue
+++ b/src/components/Controls/Color.vue
@@ -71,6 +71,7 @@ export default {
       return this.df.options;
     },
     selectedColorLabel() {
+      if(!this.colors) return this.value;
       const color = this.colors.find((c) => this.value === c.value);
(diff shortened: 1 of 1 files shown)


Isaac-GC:
Hi @shahzadbinshahjahan Can you fix the linting issue? 

Should just be able to run `yarn prettier --write src/components/Controls/Color.vue` and be good

shahzadbinshahjahan:
Hi @Isaac-GC, I have fixed the linting issue. Thank you for your help.

Issue #860, opened 2024-02-22
🐛 [Bug] - Purchase Payment Bug on Both Party


smalldale:
### Expected Behavior

Purchase Payment `From Account` should use the same `From Account` from single role party.
![image](https://github.com/frappe/books/assets/49019693/baad4712-40e0-447d-b3e0-455ba1b40ea9)
![image](https://github.com/frappe/books/assets/49019693/9c0cf3c7-98c4-455d-a69f-42e9a9fb2dc5)

### Current Behavior

By default, it uses _Piutang Dagang Dalam Negeri_, with the options _Piutang Lain-lain_ and _Piutang Dagang_. Same if choose Transfer as `Payment Method`.
![image](https://github.com/frappe/books/assets/49019693/2c358197-86ff-4957-b214-e0cade15c20f)
![image](https://github.com/frappe/books/assets/49019693/c364b02f-f651-40bd-bdd8-ea030b6fac88)

### Steps to Reproduce

1. Create New Company. Add 2 Suppliers, and an Item.
I create a company like the screenshot below. 
![image](https://github.com/frappe/books/assets/49019693/a1148942-20b6-4c44-9219-90d32a6ab10b)
Create 2 party, one with only supplier role and other with both roles.
For Item is Pencil like the screenshot below.
![image](https://github.com/frappe/books/assets/49019693/58f95975-f915-4e32-bb98-cde7094a860f)
3. Create a new purchase invoice for both role like this, submit it, then create new payment.
![image](https://github.com/frappe/books/assets/49019693/2c358197-86ff-4957-b214-e0cade15c20f)

As comparison, create a new purchase invoice for single role (supplier), submit it, then create new payment.

### FrappeBooks Version

from 0.20.0 to 0.21.2

### Path or Feature name

Purchase Invoice

### Country

ID

### Language

EN

### OS

Windows 10

Sutiksh27:
Hello, I am trying to solve this bug, just wanted to make sure, the From account that we can see in the payment section side should be similar to the account we have chosen while creating the single role party or both role party, right?

Abhi135721:
Hi, I am trying to resolve this bug as well. I think that is the requirement @Sutiksh27. 

I found out that when a party is created with both role and when we try to add payment, the payment type is marked as Receive even if it is a Purchase invoice. Is it expected behavior or do we need to change the payment type based on whether it is a Purchase/Sales invoice?

Sutiksh27:
[...]

PR #864, opened 2024-03-01
Fix issue : Purchase Payment Bug on Both Party #860

Abhi135721:
I found out that when a party is created with both role and when we try to add payment, the payment type is marked as Receive even if it is a Purchase invoice. This is happening in case outstanding amount is negative and when it is purchase invoice, payment type is set to receive. Changed it in Payment.ts. I am not sure if this disturbs any other flow. 

Please review and let me know.

Thanks.


Isaac-GC:
+Sutiksh27

PR #865, opened 2024-03-01
Changed Payment Reference Type

Sutiksh27:
Reasoning: If the outstanding amount is negative means that amount was deducted/debited from the account hence the reference type should be PurchaseInvoice instead of SalesInvoice.

Result - The "FromAccount" in both role party is displaying the same menu as the single role party.
Screenshot 2024-03-01 130526
Screenshot 2024-03-01 130617

Test Supplier 2 has "both" as role

### DIFF
--- a/models/baseModels/Payment/Payment.ts
+++ b/models/baseModels/Payment/Payment.ts
@@ -521,7 +521,6 @@ export class Payment extends Transactional {
     if (!party || party.role === 'Both') {
       return null;
     }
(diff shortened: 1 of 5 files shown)


Isaac-GC:
Hi @Sutiksh27 --> A PR was merged for the same issue https://github.com/frappe/books/pull/865

Closing this PR, but gonna add your name to the other one as well

PR #922, opened 2024-08-10
fix: checks doesnt start due to unusable runner

akshayitzme:
Checks are waiting forever to be started due to the runner `macos-11` seems to be no longer usable. Switching to `macos-latest` fixes the issue.


akshayitzme:
had to go with `macos-12` for the [build action](/frappe/books/blob/master/.github/workflows/build.yml) cause `better-sqlite3` seems to be unsupported with `macos-latest` runner.

PR #929, opened 2024-08-19
feat: lead

ableksaju:
A lead is a potential customer who has shown interest in your products. Leads eventually make a purchase. Additionally, you can create a sales invoice and customer record against a lead, further streamlining the sales process. 

image


akshayitzme:
LGTM 👍🏻

PR #972, opened 2024-10-09
fix: adjust 'this month' cash flow option to start from current month

### LINKED ISSUE #884: 🐛 [General Question] `This Month` in dashboard should show summary for currentDate - startDateOfMonth
### Summary

Currently, when we choose the `This Month` option in Dashboard, it shows summary between the dates currentDate and `currentDate - 1 month`. For example, it shows summary between 6-April-2024 and 6-March-2024.

I think `This Month` should show summary only for the current month as `This` refers to the current month and not the last one month period. In the above example, it will be summary between 6-April-2024 and 1-April-2024.

I think there are two possible solutions:
- If we want to show for last 1 month, we could rename the option like `1 Month`.
- We could compute `fromDate` to use current month's beginning date. In this case, this line will be:
https://github.com/frappe/books/blob/09fb5570d2c942afbdf5db8f23a25dc2be721375/src/utils/misc.ts#L32
`fromDate = toDate.startOf('month');`

ableksaju:


akshayitzme:
LGTM 👍🏻

PR #1000, opened 2024-10-24
feat: POS modern UI 

ableksaju:
image

<br/>
<br/>

image


src/components/POS/Classic/ItemsTable.vue
```diff
@@ -29,7 +29,10 @@
     </div>
   </Row>
 
-  <div class="overflow-y-auto" style="height: 80vh">
+  <div
+    class="overflow-y-auto custom-scroll custom-scroll-thumb2"
+    style="height: calc(70vh)"
```
akshayitzme: using `calc` is unnecessary here unless you're performing a calculation like `calc(2em * 5);`

src/pages/POS/LoyaltyprogramModal.vue
akshayitzme, line 1: The filename should be `LoyaltyProgramModal.vue`. Please change the import syntax as well.

src/pages/POS/POS.vue
```diff
@@ -1,6 +1,12 @@
 <template>
-  <div class="">
-    <PageHeader :title="t`Point of Sale`">
+  <div class="flex-col">
+    <PageHeader
+      :title="
+        fyo.singles.POSSettings?.posUI === 'Classic'
+          ? t`Point of Sale (Classic)`
+          : t`Point of Sale (Modern)`
```
akshayitzme: headers should be same `Point of Sale`

akshayitzme:
Validation on Discount Amount & Percentage is not working on POS. I can set a discount percentage that is > 100.

PR #1040, opened 2024-12-05
 feat: shortcuts in POS

ableksaju:


src/components/ShortcutsHelper.vue
```diff
@@ -191,6 +191,49 @@ export default defineComponent({
           },
         ],
       },
+      {
+        label: t`POS`,
```
akshayitzme: Change it to `Point of Sale`

PR #1046, opened 2024-12-10
feat: update dutch translations for accuracy

stxm:
translations are a bit more clear and
The John Doe name is replaced as per Dutch government ID card specimen


stxm:
Had to fix file encoding!

PR #1062, opened 2024-12-20
feat: option to display time in print templates

ableksaju:


src/utils/printTemplates.ts
```diff
@@ -69,9 +69,30 @@ export async function getPrintTemplatePropValues(
...
+  if (printSettings.displayTime) {
+    (values.doc as PrintTemplateData).time = getTime(doc.date as string);
+  }
+
   return values;
 }
 
+function getDate(dateString: string): string {
+  const date = new Date(dateString);
+  date.setMonth(date.getMonth() - 1);
```
mincerray1, line 83: hello, why is there a "- 1"? is that necessary? the dates in the print templates now show the previous month

akshayitzme: resolved in #1082

PR #1065, opened 2024-12-23
feat: submit and hold invoice for later use in POS

ableksaju:


src/pages/POS/POS.vue
```diff
@@ -571,13 +582,26 @@ export default defineComponent({
       await this.applyPricingRule();
       await this.sinvDoc.runFormulas();
     },
-
-    async createTransaction(shouldPrint = false) {
+    async createTransaction(shouldPrint = false, isPay = false) {
```
akshayitzme, line 585: `isPay` should be renamed to `makePayment` or `createPayment` or something else that suits the action that's performed.

PR #1110, opened 2025-01-31
feat: print template for payment entries

### LINKED ISSUE #1067: 💡 [Feature Request] - I need print option for the payment entries
### Summary

I am using frappe books for our institute, Some students' financial condition does not allow them to full payment so we need to generate partial payment receipt.

but there is not a print button to do payment.

### What problem are you trying to solve?

This can solve partial payment receipts and balance amounts to customers. On print template of reciept.

### Basic Example

I am using frappe books for our institute, Some students' financial condition does not allow them to full payment so we need to generate partial payment receipt.

### Drawbacks

none

### Reference Issues

_No response_

ableksaju:
fixes #1067

### DIFF
--- a/templates/Business.Payment.template.html
+++ b/templates/Business.Payment.template.html
@@ -72,7 +72,9 @@
   </section>
 
   <section class="w-1/2 px-12 pb-6 text-lg" v-if="doc.taxes.length">
-    <p class="flex justify-center font-semibold text-lg mt-2">{{t`Tax Summary`}}</p>
+    <p class="flex justify-center font-semibold text-lg mt-2">
+      {{t`Tax Summary`}}
+    </p>
 
     <div
       class="flex justify-between pl-5 text-base"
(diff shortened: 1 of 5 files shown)


templates/Business.Payment.template.html
```diff
@@ -0,0 +1,124 @@
...
+      <div class="col-span-2 row-start-2 font-bold">{{t`Payment Method`}}</div>
+      <div class="col-start-3 row-start-2">:</div>
+      <div class="col-span-4 col-start-4 row-start-2">
+        {{ doc.paymentMethod }}
+      </div>
+    </div>
+  </section>
+
+  <section class="w-1/2 px-12 pb-6 text-lg" v-if="doc.taxes.length">
+    <p class="flex justify-center font-semibold text-lg mt-2">Tax Summary</p>
```
akshayitzme: should be translatable

templates/Business.Payment.template.html
```diff
@@ -0,0 +1,124 @@
...
+        <div class="col-span-4 col-start-4 row-start-2 font-bold">
+          <p>{{ doc.amountPaid }}</p>
+          <p class="font-normal">{{ doc.amountPaidInWords }}</p>
+        </div>
+      </div>
+    </section>
+
+    <!-- Invoice Terms -->
+    <section class="mt-12" v-if="doc.terms">
+      <h3 class="text-lg font-semibold">Notes</h3>
```
akshayitzme: should be translatable

PR #1146, opened 2025-03-12
fix: set return invoice amount as a negative value

ableksaju:


src/pages/POS/PaymentModal.vue
```diff
@@ -371,6 +371,15 @@ export default defineComponent({
       }
 
       if (amount) {
+        if (this.sinvDoc.isReturn) {
+          this.$emit(
+            'setPaidAmount',
+            -Math.abs(this.sinvDoc.outstandingAmount?.float)
```
akshayitzme: You can replace this line using `neg()`

akshayitzme (approved):

PR #1156, opened 2025-03-25
fix: allow selecting a default account in POS transactions

ableksaju:


models/inventory/Point of Sale/POSSettings.ts
```diff
@@ -22,6 +23,10 @@ export class POSSettings extends Doc {
       accountType: AccountTypeEnum.Cash,
       isGroup: false,
     }),
+    defaultAccount: () => ({
+      isGroup: false,
+      accountType: 'Receivable',
```
akshayitzme: Replace with ` AccountTypeEnum.Receivable`

akshayitzme (approved):

PR #1176, opened 2025-04-05
fix: restrict overriding existing company data on duplicate name creation

### LINKED ISSUE #1147: 🐛 [Bug] - Creating new company with same name overrides all the data.
### Expected Behavior

If I create new company with existing name, it should atleast warn me before overwriting db file.

### Current Behavior

I just created new company to experiment, and used similar name. I just lost all of my data. I assume there is nothing we can do about that. (If we can, please tell). Just raising this issue so that nobody else face this issue.

### Steps to Reproduce

_No response_

### FrappeBooks Version

0.28.0

### Path or Feature name

Create Company

### Country

IN

### Language

EN

### OS

Macos (Apple Silicon)

### Additional OS Info

_No response_

ableksaju:
fixes #1147

### DIFF
--- a/main/registerIpcMainActionListeners.ts
+++ b/main/registerIpcMainActionListeners.ts
@@ -57,7 +57,32 @@ export default function registerIpcMainActionListeners(main: Main) {
       const backupPath = path.join(dbsPath, 'backups');
       await fs.ensureDir(backupPath);
 
-      return path.join(dbsPath, `${companyName}.books.db`);
+      let dbFilePath = path.join(dbsPath, `${companyName}.books.db`);
+
+      if (await fs.pathExists(dbFilePath)) {
+        const option = await dialog.showMessageBox({
+          type: 'question',
+          title: 'File Exists',
+          message: `Filename already exists. Do you want to overwrite the existing file or create a new one?`,
+          buttons: ['Overwrite', 'New'],
+        });
+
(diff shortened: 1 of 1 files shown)


main/registerIpcMainActionListeners.ts
```diff
@@ -57,7 +57,32 @@ export default function registerIpcMainActionListeners(main: Main) {
...
-      return path.join(dbsPath, `${companyName}.books.db`);
+      let dbFilePath = path.join(dbsPath, `${companyName}.books.db`);
+
+      if (await fs.pathExists(dbFilePath)) {
+        const option = await dialog.showMessageBox({
+          type: 'question',
+          title: 'File Exists',
+          message: `Filename already exists. Do you want to overwrite the existing file or create a new one?`,
+          buttons: ['Overwrite', 'New'],
+        });
```
akshayitzme, line 68: Is it possible to display this message using `showDialog` Component ? 

Ref: [DatabaseSelector.vue](https://github.com/frappe/books/blob/394accfa4164a8928f3dea40bc6db59440f003ae/src/pages/DatabaseSelector.vue#L378-L390)

ableksaju: `showDialog` is a frontend component, so it isn’t available in this file

PR #1183, opened 2025-04-23
fix: improve docs filtering & Added status wise Filtering

suhailanzar:


src/components/FilterDropdown.vue
```diff
@@ -151,14 +202,14 @@ import Popover from './Popover.vue';
...
-  { label: t`Is Empty`, value: 'is null' },
-  { label: t`Is Not Empty`, value: 'is not null' },
+  { label: '=', value: '=' },
+  { label: '!=', value: '!=' },
+  { label: 'like', value: 'like' },
+  { label: 'not like', value: 'not like' },
+  { label: '>', value: '>' },
+  { label: '<', value: '<' },
+  { label: 'is null', value: 'is null' },
+  { label: 'is not null', value: 'is not null' },
```
akshayitzme: Is this change really needed?. Labels are used to make the easy for the user for daily operations. For eg, from a non-technical user perspective, `!=` doesn't directly imply what it is used for.

src/components/FilterDropdown.vue
```diff
@@ -289,22 +392,33 @@ export default defineComponent({
...
     },
+
     emitFilterChange(): void {
       const filters: Record<string, [Condition, Filter['value']]> = {};
+
       for (const { condition, value, fieldname } of this.newFilters) {
+        console.log('condition is:', condition);
+        console.log('value is:', value);
+        console.log('fieldname is:', fieldname);
+
```
ableksaju: Avoid `console.log` in production unless essential.

PR #1193, opened 2025-04-29
feat: enable POS functionality without inventory

### LINKED ISSUE #1181: 🐛 [General Question] - Use POS without inventory
### Summary

Hi all,

My use case is that I sell things like a home-cooked meal, or home-made drinks. I'd like to use the POS ; at first I thought I had a problem since it didn't show up in the menu but I understood tweaking around that you have to activate the inventory management first.

As you may understand, since I don't buy the products I sell (I make them), the inventory is no real use to me and I can't just keep adding meals to a virtual stock, it makes no sense since I make them on demand…

Is there a way to use the POS without any inventory? I tried labeling my products as "services" instead, but for the moment I didn't manage to get it working, nothing shows up in the POS if I don't add stock for my meals/drinks in the inventory.

Is there a way to get around this? **Have the POS without inventory?**

Thanks you for your time, and big up to the team providing this nice tool!

ableksaju:
image

<br/>
<br/>

Allows selection of "**Non-Inventory Items**" in **POS settings** for sales without stock tracking.

fixes #1181

### DIFF
--- a/models/baseModels/Invoice/Invoice.ts
+++ b/models/baseModels/Invoice/Invoice.ts
@@ -1073,7 +1073,9 @@ export abstract class Invoice extends Transactional {
       linkedEntries = await getLinkedEntries(sinvDoc);
     }
 
-    if (!this.stockNotTransferred) {
+    const itemVisibility = this.fyo.singles.POSSettings?.itemVisibility;
+
+    if (!this.stockNotTransferred && itemVisibility === 'Inventory Items') {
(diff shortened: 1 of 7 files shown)


FelixDz:
Hello,

Not sure if that's relevant but I suppose that with this option, POS settings should appear in the config options without having to activate the Inventory management module first, which is currently the case (at least in the latest stable release).

PR #1206, opened 2025-05-12
fix: Zero outstanding for unpaid returns & alert option to save & proceed for invoice.

Gadha2311:


models/baseModels/Invoice/Invoice.ts
```diff
@@ -195,6 +195,16 @@ export abstract class Invoice extends Transactional {
     await super.afterSubmit();
 
     if (this.isReturn) {
+      await this.fyo.db.update(this.schemaName, {
+        name: this.name as string,
+        outstandingAmount: 0,
+      });
```
ableksaju: This sets `outstandingAmount` to 0 on `return`, even if the amount is **paid**. We need to show an amount, indicating what should be returned to the customer

src/pages/POS/POS.vue
```diff
@@ -907,6 +909,25 @@ export default defineComponent({
...
+    async handleSaveAndContinue() {
+      try {
+        if (!this.sinvDoc.party || !this.sinvDoc.items?.length) {
+          showToast({
+            type: 'error',
+            message: t`Please add a customer and at least one item before saving`,
+          });
+          return;
+        }
+        await this.saveInvoiceAction();
```
ableksaju: `!this.sinvDoc.items?.length` is already cheching in the routing you have to only check is the party is available or not

```suggestion
    async handleSaveAndContinue() {
      try {
        if (!this.sinvDoc.party) {
          return  showToast({
            type: 'error',
            message: t`Please add a customer before saving`,
          });
        }
        await this.saveInvoiceAction();
```

models/baseModels/Invoice/Invoice.ts
```diff
@@ -192,8 +192,6 @@ export abstract class Invoice extends Transactional {
   }
 
   async afterSubmit() {
-    await super.afterSubmit();
-
```
ableksaju: Please retain these lines of code, as they are required

ableksaju (commented):
Update the highlighted sections

PR #1223, opened 2025-05-26
fix: payment type fix while return and returnIssue status added

suhailanzar:


models/baseModels/Invoice/Invoice.ts
```diff
@@ -195,7 +195,9 @@ export abstract class Invoice extends Transactional {
     await super.afterSubmit();
 
     if (this.isReturn) {
+      await this._updateIsItemsReturned();
```
ableksaju: Duplicated call — remove it

PR #1238, opened 2025-06-08
Add arm64 targets for the linux deb & rpm binaries.

kskarthik:


kskarthik:
this also enables to provide arm64 flatpak

PR #1248, opened 2025-06-13
fix:make select field editable

Jas1rk:


ableksaju:
image

Please rework this completely — the current PR looks incomplete

mihir-kandoi:
Closing because typing a search prefix immediately writes it as the selected document value, which fails validation for fixed options. The template also contains unfinished placeholder text. Search text needs to remain separate from the selected value.

PR #1253, opened 2025-06-19
fix: Apply item discount when required quantity matches pricing rule

Gadha2311:


src/components/POS/Classic/SelectedItemRow.vue
```diff
@@ -382,14 +382,21 @@ export default defineComponent({
...
+      const hasManualDiscount = this.row.setItemDiscountAmount;
+      const isPercentageDiscount =
+        !hasManualDiscount && this.row.itemDiscountPercent !== 0;
+      const manualDiscountAmount = this.row.itemDiscountAmount;
+      const manualDiscountPercent = this.row.itemDiscountPercent;
+      if (!this.row.isReturn && quantity <= 0) {
+        if (quantity <= 0) {
+          showToast({
+            type: 'error',
+            message: 'Quantity must be greater than zero.',
```
ableksaju: `quantity <= 0` is already checked in the outer condition

PR #1271, opened 2025-07-07
fix:add pagination support for return sales invoice list

Gadha2311:


src/pages/POS/ReturnSalesInvoiceModal.vue
```diff
@@ -82,6 +82,14 @@
       </Row>
     </div>
 
+    <div class="mt-4">
+      <Paginator
+        :item-count="filteredInvoices.length"
+        :allowed-counts="[20, 40, -1]"
```
ableksaju: Reduce top & bottom margin

PR #1272, opened 2025-07-07
fix: resolve issue in grand total calculation with loyalty points applied

Gadha2311:


models/baseModels/Invoice/Invoice.ts
```diff
@@ -888,21 +888,31 @@ export abstract class Invoice extends Transactional {
   }
 
   async getLPAddedBaseGrandTotal() {
-    if (!this.initialGrandTotal) {
-      this.initialGrandTotal = this.grandTotal;
+    const totalDiscount = this.getTotalDiscount();
+
+    let baseTotal = this.fyo.pesa(0);
+    if (!this.taxes!.length) {
```
akshayitzme: If you check the existence of `this.taxes` here then you can avoid the `as Doc` assertion below. 

```ts
    if (!this.taxes || !this.taxes.length) {
```


~baseTotal = ((this.taxes ?? []) as Doc[])~

```ts
 } else {
      baseTotal = this.taxes
        .map((doc) => doc.amount as Money)
      ...
```

models/baseModels/Invoice/Invoice.ts
```diff
@@ -888,21 +888,31 @@ export abstract class Invoice extends Transactional {
...
-    const totalLotaltyAmount = await getAddedLPWithGrandTotal(
+    const totalLoyaltyAmount = await getAddedLPWithGrandTotal(
       this.fyo,
       this.loyaltyProgram as string,
       this.loyaltyPoints as number
     );
 
     if (this.isReturn) {
-      return this.grandTotal;
+      return baseTotal.abs().sub(totalLoyaltyAmount);
```
ableksaju: This logic was breaking for loyalty + discounted return invoices


```suggestion
      return this.grandTotal;
```

PR #1314, opened 2025-08-05
fix: resolve batch-wise return issue for loyalty points

Gadha2311:


models/baseModels/Invoice/Invoice.ts
```diff
@@ -822,19 +834,26 @@ export abstract class Invoice extends Transactional {
...
+
     let sinvDoc;
-    if (doc?.returnAgainst) {
+    try {
       sinvDoc = await this.fyo.doc.getDoc(
         ModelNameEnum.SalesInvoice,
         doc.returnAgainst
       );
+    } catch (error) {
+      return;
```
ableksaju: Remove try-catch – already handled in parent

models/baseModels/Invoice/Invoice.ts
```diff
@@ -1629,18 +1665,31 @@ export abstract class Invoice extends Transactional {
...
           roundFreeItemQuantity,
-          pricingRuleDoc.roundingMethod as 'round' | 'floor' | 'ceil'
+          'floor'
+        );
+      }
+
+      if (roundFreeItemQuantity <= 0) {
+        throw new ValidationError(
+          t`Cannot add free item - calculated quantity is zero or negative`
         );
```
ableksaju:
```suggestion
        throw new ValidationError(
          t`Free item "${pricingRuleDoc.freeItem as string}" was not added due to zero
           quantity`
        );
```

PR #1317, opened 2025-08-07
feat: include enabled feature data in error reports

### LINKED ISSUE #1160: 💡 [Feature Request] - Error reporting should include the data of enabled features
### Reason
People often raise issues in a limited context and sometimes they don't provide additional contexts. This makes it difficult for the developers to reason the issue. 

### What problem are you trying to solve?

_

### Basic Example

<details>
<summary><code>[show/hide]</code></summary>

```json
{
 "enableInventory": true,
 "enablePriceList": false,
}
```
</details>

### Drawbacks

_

### Reference Issues

_

Gadha2311:
#1336

### DIFF
--- a/src/errorHandling.ts
+++ b/src/errorHandling.ts
@@ -192,6 +192,79 @@ export function getErrorHandledSync<T extends (...args: any[]) => any>(
   };
 }
 
+function getFeatureFlags(): string[] {
+  const getBooleanFields = (docName: string) => {
+    const doc = fyo.singles[docName];
+    if (!doc) {
+      return {};
+    }
+
+    const schema = fyo.schemaMap[docName];
+    const booleanFields: Record<string, boolean> = {};
+
+    for (const [key, value] of Object.entries(doc)) {
+      const field = schema?.fields.find((f) => f.fieldname === key);
+      if (
+        typeof value === 'boolean' &&
+        !field?.hidden &&
+        !key.startsWith('_')
+      ) {
+        booleanFields[key] = value;
+      }
+    }
+
+    return booleanFields;
+  };
+
+  const accountingFlags = getBooleanFields('AccountingSettings');
+  const posFlags = getBooleanFields('POSSettings');
+  const inventoryFlags = getBooleanFields('InventorySettings');
+
+  const sections = [];
+
+  if (Object.keys(accountingFlags).length > 0) {
+    sections.push(
+      '**Accounting Settings**:',
+      '```json',
+      JSON.stringify(accountingFlags, null, 2),
+      '```',
+      ''
+    );
+  }
+
+  if (Object.keys(posFlags).length > 0) {
+    sections.push(
+      '**POS Settings**:',
+      '```json',
+      JSON.stringify(posFlags, null, 2),
+      '```',
+      ''
+    );
+  }
+
+  if (Object.keys(inventoryFlags).length > 0) {
+    sections.push(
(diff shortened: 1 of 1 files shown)


src/errorHandling.ts
```diff
@@ -192,6 +192,52 @@ export function getErrorHandledSync<T extends (...args: any[]) => any>(
...
 
+function getFeatureFlags(): string[] {
+  const getBooleanFields = (docName: string) => {
+    const doc = fyo.singles[docName];
+    if (!doc) return {};
+
+    return Object.entries(doc).reduce((acc, [key, value]) => {
+      const field = fyo.schemaMap[docName]?.fields.find(
+        (f) => f.fieldname === key
+      );
```
akshayitzme: Here, `find()`  runs once per key. If the number of fields is large, this creates O(n^2) complexity. 

You can fix this by building a Map of (fieldname, field) before the loop 

eg: 
```ts
const fieldsArray = fyo.schemaMap[docName]?.fields ?? [];
const fieldsMap = new Map(fieldsArray.map(f => [f.fieldname, f]));

```

and you can access the items from the map easily like, 
```ts
const field = fieldsMap.get(key);
```

[Map Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

src/errorHandling.ts
```diff
@@ -192,6 +192,52 @@ export function getErrorHandledSync<T extends (...args: any[]) => any>(
...
+  ]
+    .filter(({ flags }) => Object.keys(flags).length > 0)
+    .map(({ name, flags }) => [
+      `**${name} Settings**:`,
+      '```json',
+      JSON.stringify(flags, null, 2),
+      '```',
+      '',
+    ])
+    .flat();
```
akshayitzme: You can use `flatMap` here instead of using separate `map()` and `flat()` methods.

[FlatMap Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

src/errorHandling.ts
```diff
@@ -192,6 +192,52 @@ export function getErrorHandledSync<T extends (...args: any[]) => any>(
...
+        acc[key] = value;
+      }
+      return acc;
+    }, {} as Record<string, boolean>);
+  };
+
+  const sections = [
+    { name: 'Accounting', flags: getBooleanFields('AccountingSettings') },
+    { name: 'POS', flags: getBooleanFields('POSSettings') },
+    { name: 'Inventory', flags: getBooleanFields('InventorySettings') },
```
akshayitzme: Please use `ModelNameEnum` instead of hardcoding model names.

src/errorHandling.ts
```diff
@@ -192,6 +193,59 @@ export function getErrorHandledSync<T extends (...args: any[]) => any>(
   };
 }
 
+function getFeatureFlags(): string[] {
+  const getBooleanFields = (docName: string) => {
+    const doc = fyo.singles[docName];
+    if (!doc) return {};
+
+    return Object.entries(doc).reduce((acc, [key, value]) => {
+      const fieldsArray = fyo.schemaMap[docName]?.fields ?? [];
```
ableksaju:
```suggestion
    return Object.entries(doc as Doc).reduce((acc, [key, value]) => {
      const fieldsArray = fyo.schemaMap[docName]?.fields ?? [];
```

The condition can be removed

PR #1327, opened 2025-09-01
Add and Improve Arabic translations

osama1998H:
Enhance Arabic translations across the codebase


osama1998H:
@ableksaju @taha2002 @Faridget 
Please review my work and let me know if there are any changes that need to be made. This is crucial for my current work because the users only use the Arabic language.

PR #1332, opened 2025-09-08
fix: default account mandatory error

Gadha2311:


models/inventory/helpers.ts
```diff
@@ -15,6 +15,9 @@ import type { SerialNumberStatus } from './types';
 export async function validateBatch(
   doc: StockMovement | StockTransfer | Invoice
 ) {
+  if (doc.schemaName === 'SalesQuote') {
+    return;
+  }
```
ableksaju: Use `ModelNameEnum.SalesQuote `instead of a hardcoded string

models/inventory/helpers.ts
```diff
@@ -52,6 +55,9 @@ async function validateItemRowBatch(
 }
 
 export async function validateSerialNumber(doc: StockMovement | StockTransfer) {
+  if (doc.schemaName === 'SalesQuote') {
+    return;
+  }
```
ableksaju: Use `ModelNameEnum.SalesQuote `instead of a hardcoded string

models/inventory/helpers.ts
```diff
@@ -64,6 +70,9 @@ export async function validateSerialNumber(doc: StockMovement | StockTransfer) {
 async function validateItemRowSerialNumber(
   row: StockMovementItem | StockTransferItem
 ) {
+  if (row.parentdoc?.schemaName === 'SalesQuote') {
+    return;
+  }
```
ableksaju: Use `ModelNameEnum.SalesQuote `instead of a hardcoded string

src/components/POS/Classic/SelectedItemRow.vue
```diff
@@ -213,6 +227,7 @@
       class="pl-6 px-4 pt-6 col-span-2"
     >
       <Link
+        ref="batchField"
```
ableksaju: Not used, can be removed

models/baseModels/Item/Item.ts
```diff
@@ -24,7 +24,7 @@ export class Item extends Doc {
   itemCode?: string;
   trackItem?: boolean;
   itemType?: 'Product' | 'Service';
-  for?: 'Purchases' | 'Sales' | 'Both';
+  for?: 'Purchases' | 'Sales' | 'Both' = 'Both';
```
ableksaju: Default value 'Both' is already defined in schema, redundant here

PR #1359, opened 2025-10-01
fix :filter stock vs non-stock items when Item Visibility changes

Gadha2311:


schemas/app/inventory/InventorySettings.json
```diff
@@ -12,6 +12,24 @@
...
+        },
+        {
+          "value": "Non-Inventory Items",
+          "label": "Non-Inventory Items"
+        }
+      ],
+      "default": "Non-Inventory Items",
+      "required": true,
+      "section": "Default"
+    },
```
akshayitzme, line 22: This could have been a `Check` field @Gadha2311 . That will be easier in developing too.

models/baseModels/InvoiceItem/InvoiceItem.ts
```diff
@@ -636,7 +637,40 @@ export abstract class InvoiceItem extends Doc {
...
+
+      if (invItemVisibility === 'Inventory Items') {
+        return { ...baseFilter, trackItem: true };
+      }
+
+      if (invItemVisibility === 'Non-Inventory Items') {
+        return { ...baseFilter, trackItem: false };
+      }
+
+      return baseFilter;
```
akshayitzme, line 671: We could get better readability if `invItemVisibility` was a boolean field @Gadha2311

akshayitzme (changes requested):
Please take a look at the comments above. Ping me if you need more clarity.  

Let's use `Check` fields wherever there is only binary values. That will be easier in development, readability and debugging too.

mihir-kandoi:
Closing because the new default hides stock items from ordinary invoices, while enabling it hides service items. This shared filter prevents normal mixed stock/service selection. The filter needs an invoice-context check and an option to show both types.

PR #1362, opened 2025-10-06
feat: add update button next to version

Gadha2311:
In the System Settings page, a button named “Update” has been added. When the user clicks the Update button, a modal appears displaying the message “Update Downloaded — Restart Frappe Books to install update?” with two options, Yes and No. If the user clicks Yes, the system will automatically update to the latest version.

Screenshot from 2025-10-21 11-07-39

### DIFF
--- a/main/preload.ts
+++ b/main/preload.ts
@@ -64,7 +64,9 @@ const ipc = {
   },
 
   async downloadUpdate(): Promise<unknown> {
-    const result: unknown = await ipcRenderer.invoke('download-update');
+    const result: unknown = await ipcRenderer.invoke(
+      IPC_ACTIONS.DOWNLOAD_UPDATE
+    );
     return result;
   },
 
@@ -144,7 +146,7 @@ const ipc = {
 
   async downloadUpdateManual(): Promise<{ path: string }> {
     const result: { path: string } = (await ipcRenderer.invoke(
-      'download-update-manual'
+      IPC_ACTIONS.DOWNLOAD_UPDATE_MANUAL
     )) as { path: string };
     return result;
   },
--- a/main/preload.ts
+++ b/main/preload.ts
@@ -4,12 +4,14 @@ import type {
   SaveDialogOptions,
   SaveDialogReturnValue,
 } from 'electron';
-import { contextBridge, ipcRenderer } from 'electron';
+import { contextBridge, ipcRenderer, IpcRendererEvent } from 'electron';
 import type { ConfigMap } from 'fyo/core/types';
 import config from 'utils/config';
 import type { DatabaseMethod } from 'utils/db/types';
 import type { BackendResponse } from 'utils/ipc/types';
 import { IPC_ACTIONS, IPC_CHANNELS, IPC_MESSAGES } from 'utils/messages';
+import type { UpdateCheckResult } from 'src/utils/types';
+
 import type {
   ConfigFilesWithModified,
   Creds,
@@ -45,6 +47,25 @@ const ipc = {
         }
       );
     });
+  },
+
+  on(channel: string, callback: (...args: unknown[]) => void) {
+    const subscription = (_event: IpcRendererEvent, ...args: unknown[]) =>
+      callback(...args);
+    ipcRenderer.on(channel, subscription);
+
+    return () => {
+      ipcRenderer.removeListener(channel, subscription);
+    };
+  },
+
+  removeAllListeners(channel: string) {
+    ipcRenderer.removeAllListeners(channel);
+  },
+
+  async downloadUpdate(): Promise<unknown> {
+    const result: unknown = await ipcRenderer.invoke('download-update');
+    return result;
   },
 
   isFullscreen() {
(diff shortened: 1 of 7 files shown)


main/preload.ts
```diff
@@ -47,6 +49,25 @@ const ipc = {
...
+      ipcRenderer.removeListener(channel, subscription);
+    };
+  },
+
+  removeAllListeners(channel: string) {
+    ipcRenderer.removeAllListeners(channel);
+  },
+
+  async downloadUpdate(): Promise<unknown> {
+    const result: unknown = await ipcRenderer.invoke('download-update');
```
akshayitzme: Please set the value of `download-update`  in the `IPC_ACTIONS` enum and use it here. 

You can see other references of its usage if you could lookup this keyword `IPC_ACTIONS`

main/preload.ts
```diff
@@ -114,8 +135,18 @@ const ipc = {
...
+  async checkForUpdates(): Promise<UpdateCheckResult> {
+    const result: UpdateCheckResult = (await ipcRenderer.invoke(
+      IPC_ACTIONS.CHECK_FOR_UPDATES
+    )) as UpdateCheckResult;
+    return result;
+  },
+
+  async downloadUpdateManual(): Promise<{ path: string }> {
+    const result: { path: string } = (await ipcRenderer.invoke(
+      'download-update-manual'
```
akshayitzme: Here too, Please move the action name `download-update-manual` to enum and use it here.

main/registerAutoUpdaterListeners.ts
```diff
@@ -21,34 +23,44 @@ export default function registerAutoUpdaterListeners(main: Main) {
...
+      const isCurrentBeta = currentVersion.includes('beta');
+      const isNextBeta = nextVersion.includes('beta');
+
+      let downloadUpdate = true;
+      if (!isCurrentBeta && isNextBeta) {
+        const option = await dialog.showMessageBox({
+          type: 'info',
+          title: 'Update Available',
+          message: `Download version ${nextVersion}?`,
+          buttons: ['Yes', 'No'],
```
akshayitzme: Haven't checked out how this looks in the Ui. If you could attach a screenshot that would be helpful :). 

If the design permits rendering a longer message, Please convey the message what actions are performed if user proceeds. 

The title wrote here is `Update Available`. But the description only mentions about downloading a version. So, from a user POV,  I'm slightly confused, proceeding with Yes will download the latest version and update the software right ?. If so that should be mentioned clearly. 

Also it would be helpful to notify the user, taking a DB Backup is encouraged.

main/registerIpcMainActionListeners.ts
```diff
@@ -153,19 +154,38 @@ export default function registerIpcMainActionListeners(main: Main) {
...
+        const isCurrentBeta = currentVersion.includes('beta');
+        const isNextBeta = availableUpdate.version.includes('beta');
+
+        return {
+          available: true,
+          currentVersion,
+          nextVersion: availableUpdate.version,
+          isCurrentBeta,
+          isNextBeta,
+          releaseNotes: availableUpdate.releaseNotes,
```
akshayitzme: I see you are getting the release notes. Is it shown to the user as well ?

main/registerIpcMainActionListeners.ts
```diff
@@ -180,6 +200,37 @@ export default function registerIpcMainActionListeners(main: Main) {
...
+
+    let downloadUpdate = true;
+
+    if (!isCurrentBeta && isNextBeta) {
+      const option = await dialog.showMessageBox({
+        type: 'info',
+        title: 'Update Available',
+        message: `Download version ${availableUpdate.version}?`,
+        buttons: ['Yes', 'No'],
+      });
```
akshayitzme, line 215: Please check if you could apply the above message regarding title and description can be applied here too.

utils/messages.ts
```diff
@@ -12,6 +12,8 @@ export enum IPC_MESSAGES {
   ISFULLSCREEN_MAIN_WINDOW = 'isfullscreen-main-window',
   ISFULLSCREEN_RESULT = 'isfullscreen-result',
   CLOSE_MAIN_WINDOW = 'close-main-window',
+  UPDATE_AVAILABLE = 'update-available',
+  UPDATE_DOWNLOADED = 'update-downloaded',
```
akshayitzme, line 16: I couldn't see the usage of these two additions. Could you please verify ?

src/pages/Settings/Settings.vue
```diff
@@ -283,7 +313,10 @@ export default defineComponent({
     async onValueChange(field: Field, value: DocValue): Promise<void> {
       const { fieldname } = field;
       delete this.errors[fieldname];
-
+      if (fieldname === 'updateButton') {
+        await this.handleManualUpdateCheck();
+        return;
+      }
       try {
         await this.doc?.set(fieldname, value ?? '');
```
akshayitzme, line 321: Fallback to empty string will open doors for new runtime bugs here. 

We have fields of mixed types in the Settings Page right ?. When this logic tries to set empty string value to a `boolean` it will fail.  Please revert this and see if you could find alternate solutions.

akshayitzme (changes requested):
Please take a look at the mentioned issues. 

Other than that, great going @Gadha2311 . Happy to see your improvements ✨

mihir-kandoi:
Closing because failed update checks are shown as “You are running the latest version.” The update-available handler also downloads prereleases before the previous confirmation. Error reporting and prerelease confirmation need to be preserved.

PR #1371, opened 2025-10-17
feat:  Added Remove Filter checkbox in System Settings

Gadha2311:
This update introduces a new checkbox option named “Remove Filter” in the System Settings.When enabled, this option removes all default filters applied across the system, allowing users to view and select all available options instead of restricted or pre-filtered ones.

### DIFF
--- a/models/baseModels/Payment/Payment.ts
+++ b/models/baseModels/Payment/Payment.ts
@@ -674,6 +674,12 @@ export class Payment extends Transactional {
           } else {
             return 'Receive';
           }
+        } else if (partyDoc.role === 'Both') {
+          if (refDoc?.isSales && refDoc.isReturn) {
+            return 'Pay';
+          } else {
+            return 'Receive';
+          }
         }
 
         if (outstanding?.isZero() ?? true) {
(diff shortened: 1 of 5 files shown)


models/baseModels/Payment/Payment.ts
```diff
@@ -674,6 +674,12 @@ export class Payment extends Transactional {
           } else {
             return 'Receive';
           }
+        } else if (partyDoc.role === 'Both') {
+          if (refDoc?.isSales && refDoc.isReturn) {
+            return 'Pay';
+          } else {
+            return 'Receive';
+          }
```
akshayitzme: Can we replace these magic strings with Enums like this ? 

`PartyRoleEnum.Customer` 
`PaymentTypeEnum.Pay`

src/components/Controls/Link.vue
```diff
@@ -155,7 +155,7 @@ export default {
       this.filtersDisabled = true;
       this.results = [];
       setTimeout(() => {
-        this.toggleDropdown(true);
+        this.isDropdownOpen = true;
```
akshayitzme, line 158: Any specific reason to edit this line ? Isn't `toggleDropdown` working as expected  ?

akshayitzme (changes requested):
Please take a look at the comments. Please try to avoid using magic strings much as possible. 

Why ? [Here's some detail](https://softwareengineering.stackexchange.com/a/365344)

PR #1383, opened 2025-10-30
feat : Party ledger

Gadha2311:
Added functionality to open the Ledger directly from each Party document using the View button. When selected, it navigates to the ledger report and automatically filters the data based on the selected party name


reports/LedgerReport.ts
```diff
@@ -111,10 +111,26 @@ export abstract class LedgerReport extends Report {
...
+        fields: ['name', 'role'],
+        filters: { name: ['in', uniqueParties] },
+      });
+      const roleMap = new Map(
+        partyRoles.map((p) => [p.name, p.role as string])
+      );
+      this._rawData.forEach((entry) => {
+        entry.role = roleMap.get(entry.party) || '';
+      });
+    }
```
ableksaju: Can be removed

reports/types.ts
```diff
@@ -66,6 +66,7 @@ export interface LedgerEntry {
   referenceType: string;
   referenceName: string;
   party: string;
+  role: string;
```
ableksaju: can be removed

src/pages/CommonForm/CommonForm.vue
```diff
@@ -462,7 +463,7 @@ export default defineComponent({
         return;
       }
 
-      this.doc.once('afterSync', async () => {
+      void this.doc.once('afterSync', async () => {
```
ableksaju: Can be removed

PR #1385, opened 2025-11-04
feat: Added Terms and Conditions section to print template

Gadha2311:
In the Print Settings, a checkbox is available to enable Terms and Conditions.When this option is checked, a text field appears, allowing users to enter the desired Terms and Conditions content.The entered text will then be displayed at the bottom of the last page in the printed document.

### DIFF
--- a/schemas/app/PrintSettings.json
+++ b/schemas/app/PrintSettings.json
@@ -140,16 +140,16 @@
       "section": "Customizations"
     },
     {
-      "fieldname": "termsAndConditions",
-      "label": "Terms and Conditions",
-      "fieldtype": "Text",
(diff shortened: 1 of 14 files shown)


schemas/app/PrintSettings.json
```diff
@@ -133,6 +133,18 @@
       "fieldtype": "Check",
       "section": "Customizations"
     },
+    {
+      "fieldname": "displaytermsandconditions",
+      "label": "Display Terms and Conditions",
+      "fieldtype": "Check",
+      "section": "Customizations"
+    },
```
ableksaju, line 141: Use camel case for `displaytermsandconditions `

PR #1393, opened 2025-11-19
Bank Statement Import (QIF / OFX)

tonydarcy:
Adds button to the Journal Entries Page & sidebar to import a bank statement from a QIF or OFX file. It also allows Journal Transactions to be edited if the Entry Type is "Bank Entry" to allow manual allocation of these transactions.

(Resubmitting due to noticing a bug in the original file - the description from the bank statement wasn't being added - this is now added to the User Remark Field).

Here are some screenshots:
image
image
image
image

### DIFF
--- a/models/baseModels/JournalEntry/JournalEntry.ts
+++ b/models/baseModels/JournalEntry/JournalEntry.ts
@@ -17,7 +17,6 @@ import {
 import { Transactional } from 'models/Transactional/Transactional';
 import { Money } from 'pesa';
 import { LedgerPosting } from '../../Transactional/LedgerPosting';
-import { ModelNameEnum } from 'models/types';
 
 export class JournalEntry extends Transactional {
   accounts?: Doc[];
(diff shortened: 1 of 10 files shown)


tonydarcy:
To better align with the principles of the existing program, I have now edited the code so that it creates the bank statement entries as "Not Submitted" (drafts) instead of submitted and then allowing editing of submitted Bank Entry transactions.
I have also fixed some lint issues by running 'yarn lint --fix'.

tonydarcy:
Hi Admins. I just wanted to check whether this feature is something you'd consider adding? Is there anything else you need me to do before this pull request can be merged?

mihir-kandoi:
Closing because a save failure after successful rows is returned as success, and the UI clears the remaining source transactions. The OFX parser also drops NAME/MEMO descriptions when their closing tags are omitted. Import errors and supported OFX formats need complete handling.

PR #1397, opened 2025-11-23
Add Albanian language support

xoniks:
Added Albanian ('sq') to the language code map and introduced initial Albanian translations in translations/sq.csv to support localization.


xoniks:
@ableksaju can I close this or there are anything else I need to do ?

PR #1422, opened 2026-01-12
 fix:add loyaltypoint in salesinvoice

Gadha2311:


models/baseModels/Invoice/Invoice.ts
```diff
@@ -992,6 +993,21 @@ export abstract class Invoice extends Transactional {
...
+        if (!this.party) {
+          return 0;
+        }
+
+        const partyDoc = (await this.fyo.doc.getDoc(
+          ModelNameEnum.Party,
+          this.party
+        )) as Party;
+
+        return partyDoc?.loyaltyPoints || 0;
```
ableksaju:
```suggestion
        const loyaltyPoints = await this.fyo.getValue(
          ModelNameEnum.Party,
          this.party as string,
          'loyaltyPoints'
        )

        return loyaltyPoints || 0;
```

schemas/app/ERPNextSyncSettings.json
```diff
@@ -49,7 +49,8 @@
       "label": "Sync Data From Server",
       "fieldname": "syncDataFromServer",
       "fieldtype": "Button",
-      "section": "Default"
+      "section": "Default",
+      "bold": true
```
ableksaju: Edit the button component so it doesn’t look like an input field. Add proper button click effects, and include a color option to set the button color.

PR #1441, opened 2026-02-02
fix: added integration error log

Gadha2311:


src/utils/erpnextSync.ts
```diff
@@ -539,9 +604,17 @@ export async function performInitialFullSync(fyo: Fyo) {
     docsByType[docType].push(doc);
   }
 
+  for (const docType of processOrder) {
+    if (docsByType[docType]) {
+    }
+  }
```
ableksaju: Why this loop

PR #1449, opened 2026-02-16
fix: incorrect rounding of JPY amounts (zero-decimal currency)

### LINKED ISSUE #770: bug: incorrect rounding of JPY amounts
<h2>Description</h2>

While working with JPY, I encountered an issue related to incorrect rounding of amounts. I've traced the problem, which appears to be related to how the system handles precision.

**Expected behavior**: 123.99 should be rounded down to 123
**Current behavior**: 123.99 rounded up to 124

JPY typically has no decimals, the minimal denominator is 1, and amounts with decimals are typically rounded down to the nearest whole yen. 

Setting Display Precision does help hide decimals but does not resolve the incorrect amount.

It might have something to do with `countryInfo.json` and `internalPrecision` fields on the `SystemSettings` schema. I haven't figured out how to fix it yet. Changing `currency_fraction` and `currency_fraction_units ` in the  `countryInfo.json` doesn't seem to do anything, even after creating a new Organization to instantiate a new DB. Tinkering with `internalPrecision` didn't help it either.

<h2>Steps to Reproduce</h2>

- Use JPY as the system's currency
- Create Tax: in my case, it's a 10% Consumption Tax
- Create a Sales Item with the rate that would end up with >= .5 decimal when applying your Tax Rate
- Start creating a Sales Invoice. Select the Sales Item, and you'll see that the amount is rounded up

<h2>Info</h2>

**Version**: `0.19.0`
**Platform**: `darwin`
**Path**: `/edit/SalesInvoice/New Sales Invoice 01`
**Language**: `-`
**Country**: `jp`

willneill:
## Summary

- **Bug:** JPY amounts displayed with 2 decimal places (e.g. ¥1234.99) instead of rounding to whole yen (¥1235). This affected all zero-decimal currencies.
- **Root cause 1:** `countryInfo.json` had `currency_fraction_units: 100` for Japan, but JPY has no fractional units — should be `1`.
- **Root cause 2:** `updateSystemSettings()` never set `displayPrecision` based on the currency, so it always defaulted to `2` regardless of country.

### `fixtures/countryInfo.json`
- Japan: `currency_fraction_units` changed from `100` to `1`
- Japan: `currency_fraction` changed from `"Sen[G]"` to `""` (Sen hasn't been used since 1953)
- Japan: added `smallest_currency_fraction_value: 1`

### `src/setup/setupInstance.ts`
- `updateSystemSettings()` now derives `displayPrecision` from `currency_fraction_units` using `Math.log10()`:
  - `1` → precision `0` (JPY, KRW)
  - `100` → precision `2` (USD, EUR, INR)
  - `1000` → precision `3` (BHD, KWD)

### `tests/testJPYPrecision.spec.ts` (new)
- Sets up a full instance with Japan/JPY
- Verifies `displayPrecision` is stored as `0` in SystemSettings
- Verifies `fyo.pesa(1234.99).round()` produces `'1235'` (0 decimal places)
- Verifies tax calculation on ¥1127 + 10% rounds to `'1240'`

## Test plan

- [x] New JPY-specific test suite (3 tests)
- [x] All 911 existing tests continue to pass (including India/INR setup tests)
- [x] Prettier passes

Fixes #770

### DIFF
--- a/fixtures/countryInfo.json
+++ b/fixtures/countryInfo.json
@@ -1060,8 +1060,9 @@
   "Japan": {
     "code": "jp",
     "currency": "JPY",
-    "currency_fraction": "Sen[G]",
-    "currency_fraction_units": 100,
+    "currency_fraction": "",
+    "currency_fraction_units": 1,
+    "smallest_currency_fraction_value": 1,
     "currency_name": "Yen",
     "currency_symbol": "¥",
     "timezones": ["Asia/Tokyo"],
--- a/src/setup/setupInstance.ts
+++ b/src/setup/setupInstance.ts
@@ -133,11 +133,16 @@ async function updateSystemSettings(
   const systemSettings = await fyo.doc.getDoc('SystemSettings');
   const instanceId = getRandomString();
 
+  const fractionUnits = countryOptions.currency_fraction_units ?? 100;
+  const displayPrecision =
+    fractionUnits > 0 ? Math.round(Math.log10(fractionUnits)) : 2;
+
   await systemSettings.setAndSync({
     locale,
     currency,
     instanceId,
     countryCode,
+    displayPrecision,
   });
 }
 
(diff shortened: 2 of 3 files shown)


mihir-kandoi:
Closing because precision is selected from the country instead of the company currency: Japan/USD gets zero decimals and India/JPY gets two. The patch also does not implement the linked floor-rounding request. The replacement corrects currency precision; it does not impose a tax-rounding policy.

PR #1453, opened 2026-02-16
fix: subtract discount amount from line total, not per-unit rate

### LINKED ISSUE #1277: Discount amount gets reduced in the rate of the product instead of the amount.
<h2>Description</h2>
Discount amount gets reduced in the rate of the product instead of the amount.
Tried to add a discount amount to the product in the sales invoice and it gets reduced in the rate of the product instead of the amount. Makes negative values that should not be possible.

<h2>Steps to Reproduce</h2>
Tried to add a discount amount to the product in the sales invoice and it gets reduced in the rate of the product instead of the amount. Makes negative values that should not be possible.

Image

<h2>Info</h2>

**Version**: `0.32.0`
**Platform**: `darwin`
**Path**: `/edit/SalesInvoice/New%20Sales%20Invoice%2002`
**Language**: `-`
**Country**: `in`

willneill:
## Description

Fixes #1277

`getDiscountedTotalBeforeTaxation()` in `InvoiceItem.ts` was computing the discounted total incorrectly when the discount type is set to a fixed amount (rather than a percentage).

**Before (buggy):**
```
(Rate - DiscountAmount) × Quantity
```

**After (fixed):**
```
Rate × Quantity - DiscountAmount
```

### Example

- Rate: ₹100, Quantity: 3, Discount Amount: ₹50
- **Before:** (100 - 50) × 3 = **₹150** (wrong — the ₹50 discount is effectively applied 3×)
- **After:** 100 × 3 - 50 = **₹250** (correct — flat ₹50 off the line total)

The bug only manifests when `quantity > 1`. The existing code comment on line 819 already documents the intended formula as `Quantity * Rate - DiscountAmount`, confirming this is a bug and not a design choice.

The percentage-based discount path was already correct (`rate.mul(quantity).percent(...)`) — only the fixed-amount path had the operand order reversed.

## Test

Added `testDiscount.spec.ts` that:
1. Creates a sales invoice item with rate=100, qty=3, discount amount=50
2. Asserts `itemDiscountedTotal` equals 250 (not 150)

### DIFF
--- a/models/baseModels/InvoiceItem/InvoiceItem.ts
+++ b/models/baseModels/InvoiceItem/InvoiceItem.ts
@@ -821,7 +821,7 @@ function getDiscountedTotalBeforeTaxation(
    */
 
   if (setDiscountAmount) {
-    return rate.sub(itemDiscountAmount).mul(quantity);
+    return rate.mul(quantity).sub(itemDiscountAmount);
   } else if (itemDiscountPercent > 0) {
     return rate.mul(quantity).percent(itemDiscountPercent);
   }
--- a/models/baseModels/tests/testDiscount.spec.ts
+++ b/models/baseModels/tests/testDiscount.spec.ts
@@ -1,0 +1,87 @@
+import test from 'tape';
+import { closeTestFyo, getTestFyo, setupTestFyo } from 'tests/helpers';
+import { ModelNameEnum } from 'models/types';
+import { SalesInvoice } from '../SalesInvoice/SalesInvoice';
+
+const fyo = getTestFyo();
+setupTestFyo(fyo, __filename);
+
+const itemData = {
+  name: 'Widget',
+  rate: 100,
+  unit: 'Unit',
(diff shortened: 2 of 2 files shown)


leemaria:
Thank you for your fix! I was having the same problem and found your PR. However, it looks like your fix only addresses the UI and the values being sent to the database still have the inflated discount amount. 

It looks like in `models/baseModels/Invoice/Invoice.ts`, there are 2 places where `.mul(quantity)` is still happening. Line 597 and line 649.
https://github.com/frappe/books/blob/master/models/baseModels/Invoice/Invoice.ts#L596

I'm new to open source, so I apologize if this is not the proper way to leave a suggestion.

mihir-kandoi:
Closing because changing only the line-total helper leaves invoice and posting discount calculations multiplied by quantity. For 3 units at 100 with a flat discount of 50, the line total becomes 250 while the other calculations still deduct 150.

PR #1464, opened 2026-02-16
fix: upgrade better-sqlite3 to v12 for Node 24 compatibility

willneill:
## Summary
- Upgrade `better-sqlite3` from `^9.2.2` to `^12.6.2` to fix compilation failure on Node.js 24
- Upgrade `@types/better-sqlite3` from `^7.6.4` to `^7.6.13` to match

## Context
`better-sqlite3` v9 fails to compile on Node.js 24 because V8's headers now require C++20. The old version's build system doesn't pass `-std=c++20` to the compiler, causing ~20 compilation errors. v12.6.2 has proper C++20 build support and explicitly lists Node 24 in its `engines` field. There are no API breaking changes between v9 and v12 — all major version bumps were strictly about dropping old Node/Electron runtime support.

## Test plan
- [x] `yarn install` completes successfully
- [x] `yarn build` completes successfully, producing DMG and zip outputs
- [x] Verify app launches and database operations work correctly


the-narwhal:
I've laid the ground work for upgrading frappe books from electron v22 to v41 which required similar dependency upgrades. e.g. better-sqlite3 to v12.x. 

Currently I'm testing it on all the architectures and operating systems i have access to.

similar PR: #1518

mihir-kandoi:
Closing because better-sqlite3 v12 cannot be upgraded independently while the app uses the Electron 22 runtime. Please include this update in a coordinated Electron and Node migration.

Issue #1495, opened 2026-03-26
🐛 [Bug] - Drop down in Account Creation field is cut off


simatao:
### Expected Behavior

<img width="1204" height="618" alt="Image" src="https://github.com/user-attachments/assets/3a211b58-1ba0-4999-b4ac-86e285e1b48b" />

The account type dropdown is cut off

### Steps to Reproduce

Common
Journal Entry
New Entry
Account
New Account
Select a root type
select a parent account
account type field is half hidden

### FrappeBooks Version

0.36.0

### Path or Feature name

account entry

### Country

ae

### Language

en

### OS

Windows 11

vishram2997:
Fixed at https://github.com/frappe/books/pull/1498

PR #1498, opened 2026-04-01
Bug - 1495 fixed drop down issue for enum drop down

vishram2997:


mihir-kandoi:
Closing because the fixed dropdown always opens below the control, without viewport bounds or scroll repositioning. Options remain clipped near the bottom of the window, so #1495 is only partly addressed. The menu needs viewport-aware positioning.

vishram2997:
@mihir-kandoi  with my changes it doesn't clipped near botoom
Screenshot 2026-09-07 122250

PR #1510, opened 2026-04-05
feat: configurable page margins with printer-safe defaults (report-print)

the-narwhal:
Problem
When opening the Report Print View for any financial report (General Ledger, Profit & Loss, Trial Balance, and Balance Sheet) and clicking **Save as PDF** or **Print**, the report content renders flush to the page edge. This causes two issues:
- The output looks visually unprofessional with no breathing room around the content
- Most laser and inkjet printers have a hardware-enforced minimum unprintable zone (~0.4–1.2 cm), so content near the edge is physically clipped when sent to a printer

### Solution
A **Page Margins** section has been added to the existing right-hand configuration pane in the Report Print View. It sits between the Size Selection and Pick Columns sections.

**Controls**
- Four `Float` inputs — **Top**, **Bottom**, **Left**, **Right** — each accepting 0–5 cm with decimal precision
- A **"Use Recommended"** button that sets all four sides to `1.5 cm` in one click
- A hint line explaining the recommended value: *"1.5 cm (~0.6 in) per side is a safe default for most printers."*

**Default behaviour**
Margins initialise to **1.5 cm** on all sides. This was chosen because:
| Printer type | Typical hardware minimum |
|---|---|
| Laser (HP, Canon, Brother, Xerox) | ~0.4 cm (0.17 in) |
| Inkjet | 0.3–1.2 cm |
| General safe zone | ≤ 0.6 cm |

1.5 cm comfortably clears every printer's minimum and matches the default margins used by word processors such as Microsoft Word and Google Docs.

**Why no backend changes were needed**
The margins are applied as an inline CSS `padding` on the content container div inside `ScaledContainer`. Because inline `style` attributes have higher CSS specificity than the `* { padding: 0 }` rule in the print stylesheet, the padding survives the print render unchanged. The padded HTML is embedded directly in the content string passed through `constructPrintDocument`, so both `printToPDF` (Save as PDF) and the native `print` dialog path pick it up automatically. The existing `@page { margin: 0 }` contract in the print CSS is preserved.

### Files changed
| File | Change |
|---|---|
| `src/pages/PrintView/ReportPrintView.vue` | Added `Float` import, `margins` data, `contentPaddingStyle` computed, `setRecommendedMargins` method, margin UI section in settings pane |

### Testing
1. Open any report (General Ledger, Profit & Loss, Balance Sheet, etc.)
2. Click the printer icon → Report Print View opens
3. Confirm the preview shows 1.5 cm inset content by default
4. Adjust Top/Bottom/Left/Right inputs — preview updates live
5. Click **"Use Recommended"** — all four inputs reset to 1.5
6. Click **Save as PDF** — open the PDF and confirm margins are applied
7. Click **Print** — send to a physical printer and confirm content is not clipped

Example of new margin configuration set to 0:
Screenshot 2026-04-05 at 8 44 40 AM

Example of new margin configuration set to 1.5:
Screenshot 2026-04-05 at 8 44 52 AM

### DIFF
--- a/src/pages/PrintView/ReportPrintView.vue
+++ b/src/pages/PrintView/ReportPrintView.vue
@@ -33,7 +33,7 @@
           :height="size.height"
           :show-overflow="true"
         >
-          <div class="bg-white mx-auto">
+          <div class="bg-white mx-auto" :style="contentPaddingStyle">
             <div class="p-2">
               <div class="font-semibold text-xl w-full flex justify-between">
                 <h1>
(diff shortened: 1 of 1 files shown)


mihir-kandoi:
Closing because padding one outer report block does not create top and bottom margins on each printed page. The print CSS and PDF options still use zero page margins, so multi-page reports remain exposed to clipping. The settings need to control actual page margins.

PR #1518, opened 2026-04-07
chore: upgrade Electron 22 → 41, fix all breaking API changes

### LINKED ISSUE #1516: chore: upgrade Electron from v22 (EOL) to a supported version
## Current Situation

Frappe Books currently pins **Electron `22.3.27`** in `package.json`. Electron 22 reached its **end-of-life on October 10, 2023** — that was its *extended* EOL, granted specifically to accommodate users still on Windows 7/8/8.1. The normal EOL would have been even earlier.

Since that date, Electron 22 has received no security patches, no bug fixes, and no upstream Chromium or Node.js updates from the Electron team. We are currently **~2.5 years past EOL**.

**Electron 22 ships with:**
- Chromium **108** (current: 146)
- Node.js **16.x** (current in Electron: 24.x)

This means any Chromium-level CVEs published since late 2022 are unpatched in every released build of Frappe Books.

---

## Before We Proceed — Any Known Blockers?

I want to flag this before diving in: **was there a deliberate reason this hasn't been upgraded?**

A few possibilities I'm aware of:
- **Windows 7/8/8.1 support** — Electron 23+ dropped those OS versions entirely. If we have users on those systems, upgrading means dropping them.
- **Native module compatibility** — `better-sqlite3` is a native addon and requires a rebuild for each Electron/Node.js ABI. Jumping 19 major versions is a non-trivial ABI change.
- **Upstream dependency constraints** — something in the dependency tree may have been blocking the upgrade silently.

If there's context I'm missing, please share it before work begins.

---

## Proposed Plan

The target is **Electron 41** (currently the latest stable, `41.1.1`), which ships with Chromium 146 and Node.js 24.

Given the size of the jump, I'd suggest an incremental approach to isolate breaking changes:

- [x] **22 → 28**: First major checkpoint (ESM support landed, several deprecated APIs removed)
- [x] **28 → 35**: Mid-range checkpoint (preload script API redesign, `console-message` signature change)
- [x] **35 → 41**: Final leg to current stable

At each step:
- Rebuild and verify `better-sqlite3` and any other native deps via `electron-rebuild`
- Check for removed/renamed APIs against the Electron breaking changes docs
- Run the full test suite (`yarn test`, `yarn uitest`)
- Do a smoke-test build for each platform

---

## Looking for Collaborators

This is a meaningful undertaking and I'd welcome help. If you:
- Have already looked into this and know where the pain points are
- Want to take ownership of a specific platform (Windows, macOS, Linux) for testing
- Have experience with `better-sqlite3` ABI compatibility across Electron versions
- Can help review breaking changes across the intermediate versions

...please jump in here or on the linked PR once it's open. Let's coordinate in this issue before opening multiple competing PRs.

I'll begin with the 22 → 28 leg and report back with findings.

the-narwhal:
Fixes #1516

## What changed

Electron 22 reached end-of-life on **October 10, 2023**. This PR upgrades the app to **Electron 41.1.1** (Chromium 146, Node.js 24) and resolves every breaking change between the two versions.

## Breaking API changes fixed

| File | What changed | Why |
|---|---|---|
| `main.ts` | `protocol.registerBufferProtocol` → `protocol.handle` + `net.fetch` | Removed in Electron 29 |
| `main/printHtmlDocument.ts` | `webContents.print(opts, callback)` → Promise-based form | Callback overload removed in Electron 36 |
| `main/registerAppLifecycleListeners.ts` | `app.on('ready')` → `app.whenReady()` | Eliminates ready-event race condition |
| `main/registerAppLifecycleListeners.ts` | `VUEJS3_DEVTOOLS` → `VUEJS_DEVTOOLS` | Renamed export in `electron-devtools-installer` |

## Linux title bar regression fix

Electron 41 now properly implements `titleBarStyle: 'hidden'` on Linux — in Electron 22 it was a no-op, so the native OS title bar still showed. With the upgrade, it started hiding the native title bar on Linux, but there is no custom Linux title bar component in the renderer (only Windows has one). Fixed by applying `titleBarStyle: 'default'` on Linux only. macOS and Windows behaviour is unchanged.

## Dependencies updated

| Package | Before | After | Reason |
|---|---|---|---|
| `electron` | `22.3.27` | `41.1.1` | The upgrade itself |
| `better-sqlite3` | `^9.2.2` | `^12.8.0` | v9 build config doesn't pass `-std=c++20`, which Electron 41's V8 headers explicitly require |
| `@types/better-sqlite3` | `^7.6.4` | `^7.6.12` | Align with new package version |
| `node-abi` (resolution) | `^3.54.0` | `^3.89.0` | Older version didn't know Electron 41's ABI, blocking `electron-rebuild` entirely |

## Housekeeping uncovered during the upgrade

- Added `main.ts` to `tsconfig.json` include — the root entry point was invisible to the type checker
- Removed stale `"types": ["webpack-env"]` from `tsconfig.json` — the project uses Vite; `@types/webpack-env` is not installed and was causing `tsc` to error-exit before checking anything
- Imported `RequestInit` from `node-fetch` in `registerIpcMainActionListeners` to fix a type conflict with the DOM global `RequestInit`
- Fixed `app.dock?.setIcon` optional chaining — `dock` is `undefined` at the type level on non-macOS

Bugs surfaced during testing

### Printing

Two separate print bugs were found once the API migration was in place:

**Print dialog never appeared (all platforms)**
The original rewrite wrapped `webContents.print()` in `Promise.resolve()`, which resolved immediately — before the dialog appeared — causing `printWindow.close()` to be called straight away. Fixed by wrapping the call in a proper `new Promise` and passing the callback so the window stays open until the user completes or cancels.

**Print silently does nothing on Linux**
Electron's CUPS integration enumerates zero printers on Linux (both X11 and Wayland), so the dialog resolves instantly without printing. Fixed by taking a different path on Linux: generate a PDF using the existing `saveHtmlAsPdf` flow and open it with `shell.openPath()` so the user prints from the system PDF viewer (Evince, Okular, etc.), which has full OS-level printer access. `shell.openPath` is fired without `await` to prevent the IPC reply hanging under Flatpak/XDG portals.

### Binary asset corruption in production builds

`removeBaseLeadingSlash` read every output file as UTF-8 text to strip the leading slash from the `app://` base URL. Binary assets (WOFF2 fonts, PNG images) were silently corrupted: invalid byte sequences were replaced with the UTF-8 replacement character `U+FFFD` on the round-trip read/write. Fixed by allowlisting known text extensions (`.js`, `.css`, `.html`, `.svg`, etc.) and skipping everything else.

### Window drag regions blocking clicks on Mac and Linux

On Mac and Linux, containers with `-webkit-app-region: drag` consume all mouse events before they reach JavaScript — no error is thrown, clicks simply disappear. `DatabaseSelector`, `Sidebar`, and `SetupWizard` all had interactive content (nav items, form fields, buttons) inside drag regions with no `window-no-drag` override. Fixed by adding `window-no-drag` to each interactive container, matching the pattern already used correctly in `PageHeader` and `WindowsTitleBar`.

### Telemetry errors when `log_creds.txt` is missing or incomplete

On machines where `log_creds.txt` is absent or partially written, `getUrlAndTokenString` returned partially-defined creds. This caused `encodeURI(undefined)` to produce the string `"undefined"`, resulting in 404 requests to `:6969/undefined`, and `sendError` throwing `"Only absolute URLs are supported"` from `node-fetch`. Fixed by returning empty creds early if any parsed value is falsy, and guarding `logOpened()`, `log()`, and `sendError()` against firing with no creds.

### RPM packaging broken on Fedora 44+ (rpm 6.x)

`rpm 6.0` introduced a `%mkbuilddir` phase that runs before `%install` and wipes `%{buildroot}`. `fpm 1.9.3` stages files directly into `BUILD/` and emits `%install\n# noop`, expecting those files to already be in the buildroot — `%mkbuilddir` deletes them first, so `rpmbuild` exits with "File not found". Fixed by adding `build/scripts/rpmbuild-compat.sh`, a wrapper installed as `rpmbuild` on `PATH` (ahead of the real one) during Linux builds. The wrapper hard-links the staged files to a safe location before `rpmbuild` runs, then patches the spec's `%install` to restore them after `%mkbuilddir` has created a fresh empty buildroot. Hard links make the operation near-instant with no extra disk usage.

### Dev mode forced to X11

Electron 41 on Wayland defaults to the Ozone Wayland backend in dev, which conflicted with the Linux print workaround (which relies on `shell.openPath` behaving correctly). Dev mode now passes `--ozone-platform=x11` and sets `ELECTRON_OZONE_PLATFORM_HINT=x11` to force X11 consistently. Production builds are unaffected.

### Installer icon filename casing

`electron-builder-config.mjs` referenced `installericon.ico` / `uninstallericon.ico` (all lowercase), but the actual files on disk are `installerIcon.ico` / `uninstallerIcon.ico` (camelCase). This caused silent failures on case-sensitive filesystems (Linux CI). Fixed by aligning the config to match the actual filenames.

## Pre-existing `tsc` errors (not introduced by this PR)

Two type errors exist in files this PR does not touch. They were hidden before because `tsc` was bailing on the `webpack-env` error before it ever reached them:

- `reports/ProfitAndLoss/ProfitAndLoss.ts:84` — `Object is possibly 'undefined'`
- `src/utils/db.ts:34` — `Type 'unknown' is not assignable to type 'symbol | undefined'`

These are separate issues and out of scope here.

## Testing

The production file-serving path (`protocol.handle`) **cannot be exercised in `yarn dev`** — dev mode uses the Vite server. A production build (`yarn build`) is needed to verify that path before merging. 

Based specifically on what was changed, here's a focused test list:

---

### 1. App Startup & Window Lifecycle
*Affected by: `app.whenReady()` change, `titleBarStyle` fix*
- [X] App opens cleanly with no errors in the console
- [X] Linux title bar shows with native minimize, maximize, and close buttons
- [X] Closing the window and reopening via the dock/taskbar works
- [X] Quitting the app fully terminates the process (no zombie)

---

### 2. Database — open, create, use
*Affected by: `better-sqlite3` v9 → v12 rebuilt against new ABI — this is the highest-risk change*
- [X] Create a brand new company/database from scratch
- [X] Open an **existing** `.books.db` file
- [X] Navigate between modules (Sales, Purchases, Accounting, etc.) without crashes
- [X] Create a new Invoice — save it, reopen it, confirm data persists
- [X] Record a Payment
- [X] Record a Journal Entry
- [X] Run a financial report (General Ledger, Profit & Loss, Balance Sheet) and confirm data looks correct
- [X] Check the database file on disk is not corrupted after a session (open in a SQLite viewer if in doubt)

---

### 3. Printing
*Affected by: `webContents.print()` callback → Promise rewrite — previously this hung forever if called on Electron 36+*
- [X] Open an Invoice or any printable document
- [x] Trigger **Print** (not PDF) — confirm the system print dialog opens
- [x] Complete or cancel the print and confirm the app doesn't hang or crash
- [X] Confirm the window used for printing closes after the job (or cancel)

---

### 4. PDF Export
*Affected by: `saveHtmlAsPdf` uses `printToPDF`, not the changed `print()` — lower risk but worth a check since it shares the hidden print window infrastructure*
- [X] Export a document (Invoice, report) as PDF
- [X] Open the saved PDF and confirm it looks correct

---

### 5. File Dialogs & IPC
*Affected by: `RequestInit` type alignment — runtime behaviour unchanged, but worth a smoke test*
- [X] Open a file via File → Open (open dialog works, file loads)
- [X] Save a file / export data (save dialog works, file written to disk)
- [X] Check for updates (Help menu or equivalent) — confirm it doesn't throw

---

### 6. Dev Tools (dev mode only)
*Affected by: `VUEJS_DEVTOOLS` rename in `electron-devtools-installer`*
- [X] Open DevTools (they auto-open in dev) — confirm no crash in the console related to extension loading

---

### 7. Production build — protocol handler (separate from `yarn dev`)
*Affected by: `protocol.registerBufferProtocol` → `protocol.handle` + `net.fetch` — this code path is **not exercised at all** in dev mode (dev uses the Vite server)*
- [X] Run `yarn build` and install/run the output
- [X] Confirm the app loads its UI (blank screen = protocol handler broken)
- [X] Confirm assets (CSS, fonts, icons) all load correctly — no 404s in the console

---

### 8. OS/Arch tests
*Test the above on available architectures and operating systems
- [X] Confirm above tests work on Linux (wayland) arm64
- [ ] Confirm above tests work on Linux (x11) arm64
- [x] Confirm above tests work on Linux (wayland) x86
- [x] Confirm above tests work on Linux (x11) x86
- [x] Confirm above tests work on MacOS arm64
- [x] Confirm above tests work on Windows x86

---

The highest-risk item is **2 (database)** — `better-sqlite3` is the core of the app and jumped 3 major versions. Give that the most thorough run. **7 (production protocol handler)** is the one you can't test at all in `yarn dev`, so that needs a build before the PR is merged.

### DIFF
--- a/build/scripts/build.mjs
+++ b/build/scripts/build.mjs
@@ -2,6 +2,7 @@ import vue from '@vitejs/plugin-vue';
 import builder from 'electron-builder';
 import esbuild from 'esbuild';
 import fs from 'fs-extra';
+import os from 'os';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import * as vite from 'vite';
@@ -133,6 +134,33 @@ function copyPackageJson() {
 }
 
 /**
+ * On Linux, installs an rpmbuild compatibility wrapper into a temp directory
+ * and prepends it to PATH before electron-builder runs.
+ *
+ * This fixes a breaking change in rpm 6.x (Fedora 44+) that is incompatible
+ * with the bundled fpm 1.9.3:
+ *   - rpm 6.x added a %mkbuilddir phase that wipes %{buildroot} before %install
+ *   - fpm 1.9.3 stages files into BUILD/ and uses '%install # noop', expecting
+ *     those files to already be present in the buildroot
+ *   - The wrapper patches the generated spec's %install to restore the staged
+ *     files after %mkbuilddir has created a fresh empty buildroot
+ *
+ * See: build/scripts/rpmbuild-compat.sh
+ */
+function setupRpmbuildWrapper() {
+  const wrapperSrc = path.join(dirname, 'rpmbuild-compat.sh');
+  const tempBinDir = path.join(os.tmpdir(), 'frappe-books-rpmbuild-compat');
+  const wrapperDst = path.join(tempBinDir, 'rpmbuild');
+
+  fs.ensureDirSync(tempBinDir);
+  fs.copySync(wrapperSrc, wrapperDst);
+  fs.chmodSync(wrapperDst, 0o755);
+
+  process.env.PATH = `${tempBinDir}:${process.env.PATH}`;
+  console.log(`rpmbuild-compat: wrapper installed at ${wrapperDst}`);
+}
+
+/**
  * Packages the app using electron builder.
  *
  * Note: this also handles signing and notarization if the
@@ -142,6 +170,10 @@ function copyPackageJson() {
  * are passed on as builderArgs.
  */
 async function packageApp() {
+  if (process.platform === 'linux') {
+    setupRpmbuildWrapper();
+  }
+
   const { configureBuildCommand } = await await import(
     'electron-builder/out/builder.js'
   );
--- a/build/scripts/build.mjs
+++ b/build/scripts/build.mjs
@@ -171,10 +171,28 @@ async function packageApp() {
  * @param {string} base
  */
 function removeBaseLeadingSlash(dir, base) {
+  const TEXT_EXTENSIONS = new Set([
+    '.js',
+    '.mjs',
+    '.cjs',
+    '.css',
+    '.html',
+    '.svg',
+    '.json',
+    '.map',
+    '.txt',
+    '.ts',
+  ]);
+
   for (const file of fs.readdirSync(dir)) {
     const filePath = path.join(dir, file);
     if (fs.lstatSync(filePath).isDirectory()) {
       removeBaseLeadingSlash(filePath, base);
+      continue;
+    }
+
+    const ext = path.extname(file).toLowerCase();
+    if (!TEXT_EXTENSIONS.has(ext)) {
       continue;
     }
 
--- a/build/scripts/dev.mjs
+++ b/build/scripts/dev.mjs
@@ -3,12 +3,17 @@ import esbuild from 'esbuild';
 import { $ } from 'execa';
 import path from 'path';
 import { fileURLToPath } from 'url';
+import { createRequire } from 'module';
 import { getMainProcessCommonConfig } from './helpers.mjs';
+
+const require = createRequire(import.meta.url);
+const electronBin = require('electron');
 
 process.env['ELECTRON_DISABLE_SECURITY_WARNINGS'] = 'true';
 process.env['NODE_ENV'] = 'development';
 process.env['VITE_HOST'] = '127.0.0.1';
 process.env['VITE_PORT'] = 6969;
+process.env['ELECTRON_OZONE_PLATFORM_HINT'] = 'x11';
 
 /**
  * This script does several things:
@@ -123,7 +128,7 @@ async function handleResult(result) {
 }
 
 function runElectron() {
-  const electronProcess = $$`npx electron --inspect=5858 ${path.join(
+  const electronProcess = $$`${electronBin} --inspect=5858 --ozone-platform=x11 ${path.join(
     root,
     'dist_electron',
     'dev',
--- a/build/scripts/rpmbuild-compat.sh
+++ b/build/scripts/rpmbuild-compat.sh
@@ -1,0 +1,121 @@
+#!/bin/bash
+# rpmbuild-compat.sh
+#
+# Compatibility wrapper for rpmbuild to fix the fpm 1.9.3 + rpm 6.x incompatibility.
+#
+# Problem:
+#   rpm 6.x introduced a new %mkbuilddir phase that runs *before* %install.
+#   %mkbuilddir does: rm -rf %{buildroot} && mkdir -p %{buildroot}
+#   fpm 1.9.3 stages files directly into BUILD/ (passed as --define buildroot ...)
+#   and generates a spec with '%install\n# noop', expecting the staged files to already
+#   be present in the buildroot. But %mkbuilddir deletes them first, so rpm finds nothing.
+#
+# Fix:
+#   1. Hard-link all staged files from BUILD/ into a safe directory outside of BUILD/
+#      (this must happen before rpmbuild runs, so before %mkbuilddir can wipe them).
+#   2. Patch the spec's %install section to restore the files into %{buildroot} after
+#      %mkbuilddir has created a fresh empty buildroot.
+#   Hard links are used so the operation is near-instant and uses no extra disk space.
+#
+# This wrapper is automatically installed into a temp PATH directory by build.mjs on Linux.
+
+set -e
+
+WRAPPER_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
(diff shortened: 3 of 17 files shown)


the-narwhal:
@ableksaju , @akshayitzme , and @kskarthik. This was a beast of an undertaking, but I believe I've got it all working now.

I tested on a Fedora wayland arm64 laptop, fedora wayland x86 laptop, fedora x11 x86 VM, MacOS arm laptop, and a windows VM x86. Everything seems to be in working order from what I tested.

I invite you/others to test as well.

CRD716:
@ableksaju I feel like this is probably one of the most important PRs active, if you can test and merge it. EOL libraries in accounting software is bad :(

PR #1521, opened 2026-04-11
fix: correctly migrate AccountingLedgerEntry dates for UTC+ timezones

### LINKED ISSUE #819: Profit and Loss report showing incorrect data
### Expected Behavior

_No response_

### Current Behavior

1) I have a purchase invoice date with 01/01/2023 as picture below
image
2) This is Ledger report showing the date 01/01/2023 without the problem.
image
3) When I open profit and loss report. The invoice PINV1287 amount showing on year 2022 as picture below
image
4) How I know that amount is from invoice PINV1287? I checked the **AccountingLedgerEntry** table and the date is 2022-12-31. See picture below
image
image
5) Because of this. My profit and loss report showed $7.20 and extra $120 in that amount $854.40 (correct amount is 734.40) on year 2022.

### Steps to Reproduce

_No response_

### FrappeBooks Version

0.20.0

### Path or Feature name

Profit and loss

### Country

MYR

### Language

EN

### OS

Macos (Intel)

### Additional OS Info

_No response_

the-narwhal:
**This is a draft PR — feedback, discussion, and testing are very welcome.**
>
> I ran into issue #819 and may have traced it down to the root cause. I'm not deeply familiar with all the internals of this issue, so I'd especially appreciate a second set of eyes on the patch logic and the version number chosen for the rerun patch. If you're a UTC+ timezone user who has been affected, testing against a real database with pre-existing entries would be incredibly helpful in confirming the fix works end-to-end.

**Fixes #819 — P&L report shows entries in the wrong year for UTC+ timezone users**

### Root cause

`AccountingLedgerEntry.date` is a `Datetime` field serialised by calling `Date.prototype.toISOString()`. For a UTC+ user (e.g. Malaysia, UTC+8), local midnight on `2023-01-01` is `2022-12-31T16:00:00.000Z` in UTC. SQLite stores that literal string, and the P&L filter

```/dev/null/sql.sql#L1-1
WHERE date >= '2023-01-01'
```

is a plain string comparison: `"2022-12-31T16:00:00.000Z" < "2023-01-01"` → the entry is silently placed in **2022** instead of 2023.

### What was already partially in place

`timezoneDateTimeAdjuster` (added Feb 2024 in `LedgerPosting`) correctly forces every **new** ledger entry to UTC midnight (`"2023-01-01T00:00:00.000Z"`), whose UTC date string always matches the user's intended date. The `fixLedgerDateTime` migration patch was meant to backfill existing entries with the same fix — but it contained a critical async bug.

### The broken patch

```/dev/null/broken.ts#L1-7
// forEach ignores the Promise returned by async callbacks.
// The outer await only waits for the SELECT, not the UPDATEs.
trx.forEach(async entry => {
  sourceTables.forEach(async table => {
    await knex('AccountingLedgerEntry').update(...)  // never actually awaited
  });
});
```

No exception was thrown, so the patch was recorded as `failed: false` in the `PatchRun` table and will never retry.

### Changes

| File | Change |
|---|---|
| `backend/patches/v0_21_0/fixLedgerDateTime.ts` | Rewrote with `for...of` + `await` — fixes new installs upgrading from old versions |
| `backend/patches/v0_37_0/fixLedgerDateTimeRerun.ts` | New patch, new name (`fixLedgerDateTimeRerun`) at `v0.37.0` — runs for users who already had the broken patch silently execute |
| `backend/patches/index.ts` | Registers the new patch |
| `models/Transactional/LedgerPosting.ts` | Removes orphaned `// end ugly timezone fix code` comment |

### Why a new patch name?

The patch runner only re-executes a patch if it previously recorded `failed: true`. Since the broken patch recorded `failed: false`, it will never retry under the same name. A new name (`fixLedgerDateTimeRerun`) at `v0.37.0` ensures every user whose DB version is ≤ `0.37.0` — i.e. everyone currently running the app — gets the migration applied on their next launch.

### Who is affected

Only **UTC+ timezone** users (UTC+8, UTC+5:30, etc.) with entries created before the `timezoneDateTimeAdjuster` was introduced. UTC and UTC− users were never affected because their local midnight falls on the same or a later UTC day, so the stored string already compared correctly.

### DIFF
--- a/backend/patches/index.ts
+++ b/backend/patches/index.ts
@@ -6,6 +6,7 @@ import testPatch from './testPatch';
 import updateSchemas from './updateSchemas';
 import setPaymentReferenceType from './setPaymentReferenceType';
 import fixLedgerDateTime from './v0_21_0/fixLedgerDateTime';
+import fixLedgerDateTimeRerun from './v0_37_0/fixLedgerDateTimeRerun';
 import fixItemHSNField from './fixItemHSNField';
 import createPaymentMethods from './createPaymentMethods';
 
@@ -41,6 +42,11 @@ export default [
     name: 'fixLedgerDateTime',
     version: '0.21.2',
     patch: fixLedgerDateTime,
+  },
+  {
+    name: 'fixLedgerDateTimeRerun',
+    version: '0.37.0',
+    patch: fixLedgerDateTimeRerun,
   },
   { name: 'fixItemHSNField', version: '0.24.0', patch: fixItemHSNField },
   {
--- a/backend/patches/v0_21_0/fixLedgerDateTime.ts
+++ b/backend/patches/v0_21_0/fixLedgerDateTime.ts
@@ -1,40 +1,48 @@
 import { DatabaseManager } from '../../database/manager';
 
-/* eslint-disable */
 async function execute(dm: DatabaseManager) {
+  const sourceTables = [
+    'PurchaseInvoice',
+    'SalesInvoice',
+    'JournalEntry',
+    'Payment',
+    'StockMovement',
+    'StockTransfer',
+  ];
 
-    const sourceTables = [
-        "PurchaseInvoice", 
-        "SalesInvoice", 
-        "JournalEntry",
-        "Payment", 
-        "StockMovement", 
-        "StockTransfer"
-    ];
+  const entries = (await dm.db!.knex!('AccountingLedgerEntry').select(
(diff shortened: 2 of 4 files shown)


mihir-kandoi:
Closing because the rerun matches vouchers by name without checking referenceType. A purchase and sales invoice sharing a name can cause ledger dates to be overwritten from the wrong voucher.

